PR #13, opened 2019-06-28
Make a dispatch file (base-template.sh) and dispatch instead of writting to every file.

gabyx:
Instead of writting the base-tempalte to every hook, it is written once and dispatched from each hook with all arguments.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1904.072126-304d96
+# Version: 1906.281446-176a55
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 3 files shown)


rycus86:
What would be the benefit of this change?
I'm a bit concerned how portable this is:
```
DIR="$( cd "$( dirname "$0" )" && pwd )"
${DIR}/githooks-dispatch "$0" "$@"'
```

I'm not sure if it handles symbolic links and other interesting cases robustly, but copying the template files should work here too (I believe). This change also seems to break on CI, you can see the tests at https://travis-ci.org/rycus86/githooks/builds/551981070?utm_source=github_status&utm_medium=notification

I'm interested in the problem you're trying to solve with this PR. If it's just cosmetics, I'm leaning towards closing it.

In any case, thanks for the contribution!

rycus86:
> I'm interested in the problem you're trying to solve with this PR. If it's just cosmetics, I'm leaning towards closing it.

Please let me know if we're good to close this PR!

gabyx:
I am good to close it, its only manicure.

PR #18, opened 2019-07-06
Use credentials for update

gabyx:


gabyx:
@rycus86 : There is only one test failing still: `test-rules` where do I find the entry to this test? I could not find it in the travis on my branch?

rycus86:
It's in https://github.com/rycus86/githooks/blob/master/tests/test-rules.sh which runs most of the pre-commit hooks that should have run before CI already.
And the logs for it are here for your branch: https://travis-ci.org/rycus86/githooks/jobs/555685071
Looks like `shfmt` formatting issues only if I'm reading it right (on a mobile atm).

gabyx:
Finally. 😓

gabyx:
@rycus86 : Would be cool if you could have a look at the changes, its no so much. Should we try to increase the Coverage with some more Tests?

rycus86:
> @rycus86 : Would be cool if you could have a look at the changes, its no so much. Should we try to increase the Coverage with some more Tests?

OK, let me go through it quickly. I'm not too worried about tests for now, we can sort them out later.

base-template.sh
```diff
@@ -545,7 +611,7 @@ fetch_latest_update_script() {
 #   None
 #####################################################
 read_updated_version_number() {
-    LATEST_VERSION=$(echo "$INSTALL_SCRIPT" | grep "^# Version: .*" | head -1 | sed "s/^# Version: //")
+    LATEST_VERSION=$(grep "^# Version: .*" <"$INSTALL_SCRIPT" | head -1 | sed "s/^# Version: //")
```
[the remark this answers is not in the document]
rycus86: Also `sed -E` please

cli.sh
```diff
@@ -34,6 +34,7 @@ Available commands:
     readme      Manages the Githooks README in the current repository
     ignore      Manages Githooks ignore files in the current repository
     config      Manages various Githooks configuration
+    apps        Manages script folders for apps
```
rycus86: `tools` instead of `apps`?

cli.sh
```diff
@@ -51,6 +52,56 @@ print_help_header() {
...
+#####################################################
+# Set the current install dir.
+# This script location determines it.
+#
+# Sets the ${INSTALL_DIR} variable.
+#
+# Returns:
+#   0 if suncessful, 1 if not
+#####################################################
+set_install_dir() {
```
rycus86: Let's bin `set_install_dir`, `make_abs_path` and `is_abs_path` for now, and just go with `~/.githooks/tools`

cli.sh
```diff
@@ -1270,7 +1323,7 @@ git hooks update [enable|disable]
 
     if [ "$1" != "force" ]; then
         if ! is_update_available; then
-            echo "  Githooks is already on the latest version"
+            echo "  Current Githooks version: $CURRENT_VERSION is newer than the remote version: $LATEST_VERSION"
```
rycus86: Can we revert this change? I don't think the extra information is useful for most people.

cli.sh
```diff
@@ -1297,30 +1351,75 @@ record_update_time() {
...
 #
 # Returns:
-#   1 if failed the load the script, 0 otherwise
+#   0 and "$INSTALL_DIR/apps/$1/run.sh"
+#   1 and "" otherwise
 #####################################################
-fetch_latest_install_script() {
-    DOWNLOAD_URL="https://raw.githubusercontent.com/rycus86/githooks/master/install.sh"
+get_app_run_script() {
+    if [ -d "$INSTALL_DIR" ] && [ -f "$INSTALL_DIR/apps/$1/run.sh" ]; then
```
rycus86: Same comments as in the other file.

cli.sh
```diff
@@ -1468,27 +1567,14 @@ git hooks readme [add|update]
...
-
-    else
-        echo "! Failed to fetch the latest README - needs either curl or wget"
-        return 1
-    fi
-
-    # shellcheck disable=SC2181
-    if [ $? -ne 0 ]; then
+    README_FILE="$(mktemp)"
+    if ! download_file "README.md" "$README_FILE"; then
```
rycus86: This file here needs to refer to `.githooks/README.md` not the main `README.md`

cli.sh
```diff
@@ -1897,6 +1983,126 @@ config_update_last_run() {
     fi
 }
 
+#####################################################
+# Manages the app script folders.
+#
+# Returns:
+#   1 on failure, 0 otherwise
+#####################################################
+manage_apps() {
```
rycus86: `apps` to `tools` in this function

cli.sh
```diff
@@ -1897,6 +1983,126 @@ config_update_last_run() {
...
+    if [ "$1" = "help" ]; then
+        print_help_header
+        echo "
+git hooks apps install download
```
rycus86: I'd change this to `git hooks tools register <type> <script>` where `<type>` would be `download` or similar here, and you could point to any file with `<script>` that gets copied into the global folder we want it in.

tests/exec-steps.sh
```diff
@@ -39,7 +39,9 @@ for STEP in /var/lib/tests/step-*.sh; do
     elif [ $TEST_RESULT -ne 0 ]; then
         FAILURE=$(echo "$TEST_OUTPUT" | tail -1)
         echo "! $STEP has failed with code $TEST_RESULT ($FAILURE), output:"
+        echo "--------"
```
rycus86: Are these necessary? Can you please revert?

tests/step-029.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Direct template execution: execute auto-update
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Not sure why this needs to change? I think it's good to have their version printed when available, so it's easier to see when it goes wrong. If the test is successful, their output will be hidden anyway.

tests/step-032.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Direct template execution: choose to ignore the update
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-042.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Run a single-repo install and try the auto-update
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-062.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Cli tool: run forced update
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-063.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Cli tool: run an update
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-076.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Direct template execution: choose to ignore the update (non-single)
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-077.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Direct template execution: choose to ignore the update (single)
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-080.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Cli tool: add/update README
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-094.sh
```diff
@@ -2,7 +2,7 @@
 # Test:
 #   Cli tool: run an installation
 
-if ! curl --version && ! wget --version; then
+if ! curl --version >/dev/null 2>&1 && ! wget --version >/dev/null 2>&1; then
```
rycus86: Same as above, please revert this.

tests/step-100.sh
```diff
@@ -23,7 +23,7 @@ if grep -r 'github.com/rycus86/githooks' ~/test100/; then
     exit 1
 fi
 
-# run the install, and let it intall into existing repos
+# run the install, and let it install into existing repos
```
rycus86: Thanks for this!

tests/test-rules.sh
```diff
@@ -20,7 +20,6 @@ run_pre_commit_test() {
     fi
 }
 
-run_pre_commit_test no-single-quote
```
rycus86: Put this one back please.

gabyx:
Thx for the comments:
@rycus86 : are you really keen on changing back to single quotes only  ' '  ( I can do this), do you also want the HEREDOC reverted? I find it so much more flexibel and also its much more safe to use backticks in shell scripts and only if needed refere to double quotes which are always expanding, which is not always what we want, especially for regexes etc... Just give me a yes or no, than I revert it ;)

rycus86:
If you can, then yes please. I think they make it harder to make mistakes because you need to think harder if it's going to do what you want or not. :)

Thanks again and sorry for being a bit slow with this, I think we're in very different timezones that doesn't help too much.

tests/step-029.sh
gabyx: True!

tests/exec-steps.sh
rycus86: Thanks :)

cli.sh
rycus86: SLOA?

cli.sh
rycus86: Yeah, I think it's useful when working on update related changes, but not necessarily that useful otherwise.

cli.sh
rycus86: Yeah, I suppose that makes sense. But is this something you actually need at the moment? I'd start with a simple single file for now, and add extra bits only if needed. Codewise it's not much of a difference, but too many options can get confusing to people.

cli.sh
```diff
@@ -1297,30 +1328,93 @@ record_update_time() {
...
+#   script into a variable.
+#
+# Sets the ${INSTALL_SCRIPT} variable
+#
+# Returns:
+#   1 if failed the load the script, 0 otherwise
+#####################################################
+fetch_latest_install_script() {
+    INSTALL_SCRIPT="$(mktemp)"
+    # shellcheck disable=SC2181
```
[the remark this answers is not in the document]
gabyx: Thx, refactoring error.

gabyx:
Now it should be good to go.

rycus86:
Thanks a lot for all the changes @gabyx !
It looks good to me, I think it's ready to merge. :tada: 

Can I ask you to please add `(experimental)` to the new headers in the readme files for now, then I'll try to add some tests to make sure each system we test on is happy with it, then I'll drop the experimental label.

Also, once you've done that, please squash your commits into one and force push, then I'll merge it in.
Thanks!

gabyx:
Done. Lets see if the tests work.

rycus86:
Thanks a lot again @gabyx !

PR #19, opened 2019-07-08
Add installation via core.hooksPath and setup new templatedir when running in non-interactive mode

AlexGoris-Cashforce:
We noticed this project (thanks!) and wanted to use it within our organization, however to make it better suited to use in our teams, we made 2 changes:
* When running script with the `--non-interactive` parameter, the script now continues to setup a new template directory if it can't find an existing one (which is the case in most default git installations)
* The script can now be installed with `--use-hookspath` parameter, in which case the script will make the template directory the directory configured as the `core.hooksPath` parameter in the global git configuration. This allows for easier application to existing repositories, avoiding the need to search the entire disk for existing repositories.

I've updated the `README.md` for the 2nd point, to try to instruct users which mode (templateDir vs hooksPath) to use in which situation. Let me know if you approve or if you'd like to see something changed.

Thanks again for this awesome project!

### DIFF
--- a/README.md
+++ b/README.md
@@ -197,6 +197,8 @@ There is also an option to run the install script for the repository in the curr
 $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --single
 ```
 
+You can change this setting later with the [command line helper](https://github.com/rycus86/githooks/blob/master/docs/command-line-tool.md) tool, running the `git hooks config [set|reset] single` command, which affects how future updates are run, when started from the local repository.
+
 Lastly, you have the option to install the templates to, and use them from a centralized location. You can read more about the difference between this option and default one [here](#Templates-vs-global-hooks). For this, run the command below.
 
 ```shell
@@ -208,8 +210,6 @@ By default the script will install the hooks into `~/.githooks/templates/`, opti
 ```shell
 $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --use-hookspath /home/public/.githooks
 ```
-
-You can change this setting later with the [command line helper](https://github.com/rycus86/githooks/blob/master/docs/command-line-tool.md) tool, running the `git hooks config [set|reset] single` command, which affects how future updates are run, when started from the local repository.
 
 It's possible to specify which template directory should be used, by passing the `--template-dir <dir>` paramter, where `<dir>` is the directory where you wish the templated be installed.
 
--- a/README.md
+++ b/README.md
@@ -5,9 +5,33 @@
 
 A simple Shell script to support per-repository [Git hooks](https://git-scm.com/docs/githooks), checked into the actual repository that uses them.
 
-To make this work, it creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`. When one of them executes, it will try to find matching files in the `.githooks` directory under the project root, and invoke them one-by-one.
-
 > Check out the [blog post](https://blog.viktoradam.net/2018/07/26/githooks-auto-install-hooks/) for the long read!
+
+## Templates vs global hooks
+
+This script can work in one of 2 ways:
+
+- Using the git template folder (default behaviour)
+- Using the git `core.hooksPath` variable (set by passing the `--use-hookspath` parameter to the install script)
+
+Read about the differences between these 2 approaches below
+
+In both cases, the script will make sure git finds the hook templates provided by this script.
+When one of them executes, it will try to find matching files in the `.githooks` directory under the project root, and invoke them one-by-one.
+
+### Template folder
+
+In this approach, the install script creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`.
+
+This is the recommanded approach if you want to selectively control which repositories use these scripts. The install script offers to search for repositories to which it will install the hooks, and any new repositories you clone will have these hooks configured.
+
+### Central hooks location (core.hooksPath)
+
+In this approach, the install script installs the hook templates into a centralized location (`~/.githooks/templates/` by default) and sets the global `core.hooksPath` variable to that location. Git will then, for all relevant actions, check the `core.hooksPath` location, instead of the default `$(GIT_DIR)/hooks` location.
+
+This approach works more like a 'blanket' solution, where any repository (\*) will start using the hook templates, regardless of its location.
+
+**Note**(\*): It is possible to override the behaviour for a specific repository, by setting a local `core.hooksPath` variable with value `$(GIT_DIR)/hooks/`, which will revert git back to its default behaviour for that specific repository.
 
 ## Layout and options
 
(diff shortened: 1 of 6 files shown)


AlexGoris-Cashforce:
I looked at the failed travis build, however it seems that most/all failed steps (I just spot-checked a couple) are related to docker errors (e.g. 'no such image: debian:base') Is this a known issue? I find it hard to believe that our changes caused these build failures

rycus86:
Ok, I have re-run the build now, must have been a Docker Hub flake. It flagged up some shellcheck issues + formatting: https://travis-ci.org/rycus86/githooks/jobs/555818064

Seems like you used bash in some places which would break some of the tests, and there's also an unused variable that you wanted to use maybe but then didn't?

rycus86:
Thanks for this PR!

Could you potentially pull the two changes apart and do them in one PR each? I'm totally on-board with the non-interactive change, though I think we should add an extra parameter like `--allow-new-template-dir` or something similar, so you'd consent to having that set up with this additional parameter rather than saying `Y` interactively.

For the `core.hooksPath`, is that a real concern or just trying to optimize the initial scanning away? This project is basically trying to solve the shared hooks problem, which you can also do with `core.hooksPath` but it's not necessary, since the hooks from the templates will be copied into each new Git repositories (either `git init` or `git clone`), the only longer lookup is to find the repositories that are already checked out onto the disk. It doesn't have to scan the whole disk, you can set a subfolder to start the search from.
While this step can be avoided with `core.hooksPath`, I'm not sure if it's worth the extra complexity?

AlexGoris-Cashforce:
I've pulled the non-interactive change into a separate commit.

Regarding the hooksPath changes, I tried using the script as-was, mind you we have users with repositories in multiple location over multiple drives. Also we're using windows which means we have to run the script using git-bash, which perhaps introduces some performance inssues?  But on my workstation I had left the script to scan for 30+ minutes and it was still scanning for repositories. 
We want to provide this script to our users as a way for them to quickly setup githooks so they will start using the hooks included in our repositories, asking them to run a script which takes 30+ minutes to complete is just not realistic.

That's to provide some background on why I made the change. I also tried documenting this in the README to make new users aware of the 2 options and the pro's con's of each one.

install.sh
```diff
@@ -3048,7 +3090,7 @@ setup_new_templates_folder() {
     if ! is_dry_run; then
         if mkdir -p "${TILDE_REPLACED}/hooks"; then
             # Let this one go with or without a tilde
-            git config --global init.templateDir "$USER_TEMPLATES"
+            set_githooks_directory "$USER_TEMPLATES/hooks"
```
rycus86: Why the extra `/hooks` ?

install.sh
```diff
@@ -3349,6 +3391,10 @@ install_hooks_into_repo() {
 #   None
 ############################################################
 setup_shared_hook_repositories() {
+    if [ "$USE_HOOKS_PATH" = "yes" ]; then
```
rycus86: While the template files might come from the central folder, shared repos are a different story and can be useful here - maybe I need to explain these better in the readme.
Regardless of where the hooks scripts run from (templates or `core.hooksPath`), they will look for things to execute in `.githooks/<event>/*` within the repository, __plus__ there can be other hooks coming from shared hook repos which should run for all repositories. Does this make more sense?

install.sh
```diff
@@ -3392,6 +3438,28 @@ setup_shared_hook_repositories() {
...
+# 
+# Returns:
+#   none
+############################################################
+
+set_githooks_directory(){
+    if should_use_hooksPath; then
+        mark_as_hookspath_global   
+        git config --global core.hooksPath "$1"
+        git config --global --unset init.templateDir
```
rycus86: I'm not sure we should do this. Someone could have other things in the template directory, it's not only the hooks that come from there.

install.sh
```diff
@@ -2862,18 +2890,28 @@ mark_as_single_install_repo() {
...
+############################################################
+# Marks the global githooks config to use hooksPath
+#
+# Sets the 'githooks.use.hookspath' configuration.
+#
+# Returns:
+#   None
+############################################################
+mark_as_hookspath_global() {
+    git config --global githooks.use.hookspath yes
```
rycus86: Can we rename `githooks.use.hookspath` to `githooks.use.core.hooksPath` please?

rycus86:
> I'm a bit worried about how well will this work with other parts that might assume template directories, but I'm just going to set up a test runner in Travis to run the tests with this new flag on, then we'll see (hopefully).

OK, if you rebase this branch on `master` or merge `master` into this branch, then there's a new Travis job to run the tests with the new flag: https://github.com/rycus86/githooks/commit/554495b2032384ab2652de23ca3bb1ca20e69a3a#diff-9028ab884f9d93cb10c280fe639486e7

AlexGoris-Cashforce:
ugh builds are failing again because `no such image: githooks:alpine` and similar errors :(

rycus86:
That's just the last message, the actual failures are above, within the failing test cases.
```
> Executing step-020
  :: Run an install, and let it set up a new template directory (non-tilde)
! /var/lib/tests/step-020.sh has failed with code 1 (! Githooks were not installed into a new repo), output:
...
Initialized empty Git repository in /tmp/test20/.git/
grep: /tmp/test20/.git/hooks/pre-commit: No such file or directory
! Githooks were not installed into a new repo
```

Looks like tests 007 and 020 are failing.

Also, can you please change this too:
> Please also rename it in the test-corehookspath.sh script for the test.

rycus86:
Based on the tests, it looks like this change does break something. Most test cases have a few steps failing, and the one that enables this everywhere has 22 failures. :(
Are you OK investigating what's wrong with it or would you need any help?

AlexGoris-Cashforce:
Hey @rycus86 , sorry for the late response, I was busy with a few other things.
I'd like to address the errors, but in the first place I have a really hard time finding them. I always notice the error in the end saying the docker image doesn't exist.

I was able to identify a failing test: https://travis-ci.org/rycus86/githooks/jobs/560006330#L524 but that test checks if the git hooks are installed in newly created repo's, which is exactly what this PR is trying to avoid. So how can I say that test `step-004.sh` should be skipped for the corehookspath test?

rycus86:
No problem, thanks for picking it up again! I was also unavailable recently.

Oh, I think you're right, though haven't had time to look into the tests yet.
Looks like tests `007` and `020` are failing, and as you said, they probably test something directly that wouldn't make sense with your change anymore.
If you have time and could look into adapting those two tests, that would be awesome, otherwise I'll try to have a look at them today or later this week.

install.sh
rycus86: Please see this change.

install.sh
rycus86: Please drop the extraneous `/hooks`, I believe that is what's breaking the tests.

install.sh
rycus86: Please drop this new block.

install.sh
rycus86: I still don't think we should be unsetting anything here.

uninstall.sh
```diff
@@ -226,6 +249,11 @@ git config --global --unset githooks.previous.searchdir
 git config --global --unset githooks.disable
 git config --global --unset alias.hooks
 
+if using_hooks_path; then
+    git config --global --unset githooks.use.hookspath
+    git config --global --unset core.hooksPath
```
rycus86: Maybe we should set the `githooks.use.core.hooksPath` variable to the same as `core.hooksPath` so here we could check if they match and only unset if they do. This way we could make sure we don't unset preexisting settings from the user.

README.md
```diff
@@ -106,7 +130,7 @@ To try and make things a little bit more secure, Githooks checks if any new hook
...
 
-There is a caveat worth mentioning: if a terminal *(tty)* can't be allocated, then the default action is to accept the changes or new hooks. Let me know in an issue if you strongly disagree, and you think this is a big enough risk worth having slightly worse UX instead.
+There is a caveat worth mentioning: if a terminal _(tty)_ can't be allocated, then the default action is to accept the changes or new hooks. Let me know in an issue if you strongly disagree, and you think this is a big enough risk worth having slightly worse UX instead.
```
rycus86: Please revert this change.

README.md
```diff
@@ -79,7 +103,7 @@ The supported hooks are listed below. Refer to the [Git documentation](https://g
...
+The `.ignore` files allow excluding files from being treated as a hook script. They allow _glob_ filename patterns, empty lines and comments, where the line starts with a `#` character. In the above example, one of the `.ignore` files should contain `*.md` to exclude the `pre-commit/docs.md` Markdown file. The `.githooks/.ignore` file applies to each of the hook directories, and should still define filename patterns, `*.txt` instead of `**/*.txt` for example. If there is a `.ignore` file both in the hook type folder and in `.githooks`, the files whose filename matches any pattern from either of those two files will be excluded. You can also manage `.ignore` files using the [command line helper](https://github.com/rycus86/githooks/blob/master/docs/command-line-tool.md) tool, and running `git hooks ignore <pattern>`.
```
rycus86: Please revert this change.

rycus86:
I made the changes I asked for here and fixed up the tests: https://github.com/rycus86/githooks/compare/cashforce-master?expand=1

@AlexGoris-Cashforce if you're OK with it, I'll just merge it tomorrow.

AlexGoris-Cashforce:
@rycus86  Thanks for merging!
Sorry for not addressing the open issues, I had a 2-week holiday and had to finish some other stuff before that. Is there still something you'd like me to look into, or are we good as is now?

rycus86:
No worries! Thanks for coming back here!
I think it's all good but haven't tested it myself apart from CI, so if you get a chance to do so, it would be great if you could report back how well it works -- or not. :)
Especially on Windows.

Thanks a lot!

AlexGoris-Cashforce:
We are using the master branch of our repo (cashforce/githooks) on windows computers, for several weeks without issues so far. I'll update our internal documentation for new people to install from your repo starting now.

PR #22, opened 2019-07-12
Bugfix: Fail on not available shared hook.

gabyx:
The `step-103.sh` should be failing. I wrote it because it does not in my setup.

-> Installing shared hooks and then accidentially removing them and making a commit should fail because the hooks are not there.

Solution: Do an update if the hooks are not there, or fail with an error "run git hooks shared update"

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1907.110919-c0438b
+# Version: 1907.181628-ac2e27
 
 #####################################################
 # Execute the current hook,
@@ -365,8 +365,22 @@ process_shared_hooks() {
     SHARED_REPOS_LIST="$1"
     shift
 
-    update_shared_hooks_if_appropriate
+    update_shared_hooks_if_appropriate "$@"
     execute_shared_hooks "$@" || return 1
+}
+
+#####################################################
+# Sets the SHARED_ROOT and NORMALIZED_NAME
+#   for the shared hook repo url `$1`.
+#
+# Returns:
+#   none
+#####################################################
+set_shared_root() {
+    NORMALIZED_NAME=$(echo "$1" |
+        sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
+        sed -E "s/[^a-zA-Z0-9]/_/g")
+    SHARED_ROOT=~/.githooks/shared/"$NORMALIZED_NAME"
 }
 
 #####################################################
@@ -378,7 +392,14 @@ process_shared_hooks() {
 #####################################################
 update_shared_hooks_if_appropriate() {
     # run an init/update if we are after a "git pull" or triggered manually
-    if [ "$HOOK_NAME" = "post-merge" ] || [ "$HOOK_NAME" = ".githooks.shared.trigger" ]; then
+    GIT_NULL_REF="0000000000000000000000000000000000000000"
+
+    RUN_UPDATE="false"
+    [ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
+    [ "$HOOK_NAME" = ".githooks.shared.trigger" ] && RUN_UPDATE="true"
+    [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
+
+    if [ "$RUN_UPDATE" = "true" ]; then
         # split on comma and newline
         IFS=",
         "
@@ -386,13 +407,11 @@ update_shared_hooks_if_appropriate() {
         for SHARED_REPO in $SHARED_REPOS_LIST; do
             mkdir -p ~/.githooks/shared
 
-            NORMALIZED_NAME=$(echo "$SHARED_REPO" |
-                sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                sed -E "s/[^a-zA-Z0-9]/_/g")
+            set_shared_root "$SHARED_REPO"
 
-            if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
+            if [ -d "$SHARED_ROOT/.git" ]; then
                 echo "* Updating shared hooks from: $SHARED_REPO"
-                PULL_OUTPUT=$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git pull 2>&1)
+                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Update failed, git pull output:"
@@ -400,7 +419,8 @@ update_shared_hooks_if_appropriate() {
                 fi
             else
                 echo "* Retrieving shared hooks from: $SHARED_REPO"
-                CLONE_OUTPUT=$(cd ~/.githooks/shared && git clone "$SHARED_REPO" "$NORMALIZED_NAME" 2>&1)
+                [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
+                CLONE_OUTPUT=$(git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Clone failed, git clone output:"
@@ -421,11 +441,42 @@ update_shared_hooks_if_appropriate() {
 #   1 in case a hook fails, 0 otherwise
 #####################################################
 execute_shared_hooks() {
-    for SHARED_ROOT in ~/.githooks/shared/*; do
-        REMOTE_URL=$(cd "$SHARED_ROOT" && git config --get remote.origin.url)
+    # split on comma and newline
+    IFS=",
+    "
+    for SHARED_REPO in $SHARED_REPOS_LIST; do
+
+        set_shared_root "$SHARED_REPO"
+
+        # Fail if the shared root is not available (if enabled)
+        FAIL_ON_NOT_EXISTING=$(git config --global --get githooks.failOnNotExistingSharedHooks)
+
+        if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+            if [ ! -d "$SHARED_ROOT/.git" ]; then
+                echo "! Failed to execute shared hook: $SHARED_REPO"
+                echo "! Its is not available. Do run:"
+                echo "!   \$ git hooks shared update"
+                return 1
+            fi
+        fi
+
+        # Note: GIT_DIR might be set (?bug?) (actually the case for post-checkout hook)
+        # which means we really need a `-f` to sepcify the actual config!
+        REMOTE_URL=$(cd "$SHARED_ROOT" && git config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
         ACTIVE_REPO=$(echo "$SHARED_REPOS_LIST" | grep -o "$REMOTE_URL")
         if [ "$ACTIVE_REPO" != "$REMOTE_URL" ]; then
-            continue
+            echo "! Failed to execute shared hook: $SHARED_REPO"
+            echo "! Shared root url \`$REMOTE_URL\` is not the same as "
+            echo "! the requested url \`$SHARED_REPO\`. Do run:"
+            echo "!   \$ git hooks shared purge"
+            echo "!   \$ git hooks shared update"
+
+            if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+                return 1
+            else
+                echo "!   Continue..."
+                continue
+            fi
         fi
 
         if [ -d "${SHARED_ROOT}/.githooks" ]; then
@@ -433,7 +484,9 @@ execute_shared_hooks() {
         elif [ -d "$SHARED_ROOT" ]; then
             execute_all_hooks_in "$SHARED_ROOT" "$@" || return 1
         fi
+
     done
+    unset IFS
 }
 
 #####################################################
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1907.181628-ac2e27
+# Version: 1907.221206-176cca
 
 #####################################################
 # Execute the current hook,
@@ -449,12 +449,12 @@ execute_shared_hooks() {
         set_shared_root "$SHARED_REPO"
 
         # Fail if the shared root is not available (if enabled)
-        FAIL_ON_NOT_EXISTING=$(git config --global --get githooks.failOnNotExistingSharedHooks)
+        FAIL_ON_NOT_EXISTING=$(git config --get githooks.failOnNotExistingSharedHooks)
 
-        if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
-            if [ ! -d "$SHARED_ROOT/.git" ]; then
-                echo "! Failed to execute shared hook: $SHARED_REPO"
-                echo "! Its is not available. Do run:"
+        if [ "$FAIL_ON_NOT_EXISTING" = "true" ]; then
+            if [ ! -f "$SHARED_ROOT/.git/config" ]; then
+                echo "! Failed to execute shared hooks in $SHARED_REPO"
+                echo "! It is not available. To fix, run:"
                 echo "!   \$ git hooks shared update"
                 return 1
             fi
@@ -465,16 +465,16 @@ execute_shared_hooks() {
         REMOTE_URL=$(cd "$SHARED_ROOT" && git config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
         ACTIVE_REPO=$(echo "$SHARED_REPOS_LIST" | grep -o "$REMOTE_URL")
         if [ "$ACTIVE_REPO" != "$REMOTE_URL" ]; then
-            echo "! Failed to execute shared hook: $SHARED_REPO"
-            echo "! Shared root url \`$REMOTE_URL\` is not the same as "
-            echo "! the requested url \`$SHARED_REPO\`. Do run:"
+            echo "! Failed to execute shared hooks in $SHARED_REPO"
+            echo "! The URL \`$REMOTE_URL\` is different."
+            echo "! To fix it, run:"
             echo "!   \$ git hooks shared purge"
             echo "!   \$ git hooks shared update"
 
-            if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+            if [ "$FAIL_ON_NOT_EXISTING" = "true" ]; then
                 return 1
             else
-                echo "!   Continue..."
+                echo "!   Continuing..."
                 continue
             fi
         fi
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1907.110919-c0438b
+# Version: 1907.181628-ac2e27
 
 #####################################################
 # Prints the command line help for usage and
@@ -1030,12 +1030,10 @@ list_shared_hook_repos() {
             echo "  - None"
         else
             for LIST_ITEM in $(git config --global --get githooks.shared); do
-                NORMALIZED_NAME=$(echo "$LIST_ITEM" |
-                    sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                    sed -E "s/[^a-zA-Z0-9]/_/g")
 
-                if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
-                    if [ "$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
+                set_shared_root "$LIST_ITEM"
+                if [ -d "$SHARED_ROOT/.git" ]; then
+                    if [ "$(cd "$SHARED_ROOT" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
                         LIST_ITEM_STATE="active"
                     else
                         LIST_ITEM_STATE="invalid"
@@ -1066,12 +1064,11 @@ list_shared_hook_repos() {
             SHARED_REPOS_LIST=$(grep -E "^[^#].+$" <"$(pwd)/.githooks/.shared")
 
             echo "$SHARED_REPOS_LIST" | while read -r LIST_ITEM; do
-                NORMALIZED_NAME=$(echo "$LIST_ITEM" |
-                    sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                    sed -E "s/[^a-zA-Z0-9]/_/g")
 
-                if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
-                    if [ "$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
+                set_shared_root "$LIST_ITEM"
+
+                if [ -d "$SHARED_ROOT/.git" ]; then
+                    if [ "$(cd "$SHARED_ROOT" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
                         LIST_ITEM_STATE="active"
                     else
                         LIST_ITEM_STATE="invalid"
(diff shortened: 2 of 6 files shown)


gabyx:
The thing is, a clone of a repo with a .shared file results in
```shell
git clone https://git.cyfex.com/cyfex/GitUmstellung.git Test
Cloning into 'Test'...
remote: Enumerating objects: 550, done.
remote: Counting objects: 100% (550/550), done.
remote: Compressing objects: 100% (414/414), done.
remote: Total 550 (delta 306), reused 184 (delta 116) eceiving objects:  67% (369/550), 7.89 MiB
Receiving objects: 100% (550/550), 15.50 MiB | 4.71 MiB/s, done.
Resolving deltas: 100% (306/306), done.
F:/Repositories/Test/.git/hooks/post-checkout: line 425: cd: /c/Users/gabriel.nuetzi/.githooks/shared/*: No such file or directory
```
Which should not happen, its the same error. The hooks are not there yet and should probably be installed, right?

gabyx:
There are two things:
- on clone, `post-checkout` runs which  should check if any of the `.shared` hooks need to be `cloned` if there is no existing repo in `~/.gitignore/shared/*` (maybe make the path a hash of the url??)
- on any invocation of all listed `.shared` hooks, if there is no corresponding repo in `~/.gitignore/shared/*` -> fail!

What do you think?

gabyx:
I fixed the following things:
`execute_shared_hooks`:
- if a shared hooks `$SHARED_ROOT` is not there -> fails with an error:
```shell
git commit -a -m "asda"
! Failed to execute shared hook: https://git.cyfex.com/cyfex/GitHooks-General.git
  Do run:
    $ git hooks shared update
```
- `update_shared_hooks_if_appropriate`: 
  - Also run the hook for `post-checkout` (maybe improve this) , so cloneing works flawlessly.
  - Added a safety `cksum` to the `NORMALIZED_NAME` such that each url is treated really differently. (Avoid stupid overlapping)

gabyx:
Thanks for reviewing this.

install.sh
```diff
@@ -440,19 +456,40 @@ update_shared_hooks_if_appropriate() {
...
+
+        # Fail if the shared root is not available
+        if [ ! -d "$SHARED_ROOT" ] || [ ! -d "$SHARED_ROOT/.git" ]; then
+            echo "! Failed to execute shared hook: $SHARED_REPO"
+            echo "! Its is not available. Do run:"
+            echo "!   \$ git hooks shared update"
+            return 1
+        fi
+
+        REMOTE_URL=$(cd "$SHARED_ROOT" && git config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
```
gabyx: I really dont understand why I need here `-f`
Without it and with `--show-origin` shows that in `test-103.sh` it fails, showing the wrong picked config??
The working directory is not empty and points to the right shared repo clone... I have no clue. Can you help me?

tests/step-083.sh
```diff
@@ -36,7 +36,7 @@ sh /var/lib/githooks/cli.sh shared add --local /tmp/shared/first-shared.git &&
     ! sh /var/lib/githooks/cli.sh shared list | grep "first_shared" &&
     sh /var/lib/githooks/cli.sh shared remove --local /tmp/shared/second-shared.git &&
     sh /var/lib/githooks/cli.sh shared remove /tmp/shared/third-shared.git &&
-    [ ! -f "$(pwd/.githooks/.shared)" ] ||
+    [ ! -f "$(pwd)/.githooks/.shared" ] ||
```
gabyx: Thats a bug!

tests/step-083.sh
```diff
@@ -59,6 +59,6 @@ git hooks shared add --local /tmp/shared/first-shared.git &&
     git hooks shared remove --local /tmp/shared/first-shared.git &&
     ! git hooks shared list | grep "first_shared" &&
     git hooks shared remove --local /tmp/shared/second-shared.git &&
-    git hooks shared remove /tmp/shared/second-third.git &&
-    [ ! -f "$(pwd/.githooks/.shared)" ] ||
+    git hooks shared remove /tmp/shared/third-shared.git &&
```
gabyx: Bug Number 2

tests/step-083.sh
```diff
@@ -59,6 +59,6 @@ git hooks shared add --local /tmp/shared/first-shared.git &&
...
+    [ ! -f "$(pwd)/.githooks/.shared" ] ||
```
gabyx: Dito.

install.sh
gabyx: Solved: GIT_DIR is set during post-checkout. -> Made a note...

gabyx:
Thanks for merging this. I think its important as it improves usage and safety.

rycus86:
Thanks for opening this PR! I'll try to review it as soon as I can this coming week!

rycus86:
OK, so I fixed up the tests (thanks again for pointing out the bugs) and it's still passing. Can you please try to tell me again what exactly is the problem? Let's start with a failing test case, then work from there, because this PR made some big changes already, and I'm not clear on why it was needed. Thanks!

rycus86:
I've tried your new test step-103 on latest `master` and it's working fine if you don't expect it to fail where I don't expect it to fail. :) There's some spam on the output about not finding shared hooks, but otherwise everything works as expected, commits are successful and non-existing hooks are skipped. What did you expect?

tests/step-103.sh
```diff
@@ -0,0 +1,71 @@
...
+# Remove all shared hooks
+git hooks shared purge || exit 1
+
+# Clone a new one
+echo "Cloning"
+cd /tmp || exit 1
+git clone /tmp/test102 test102-clone && cd test102-clone || exit 1
+
+# Remove all shared hooks
+git hooks shared purge || exit 1
```
rycus86: This fails because you already purged, so there's nothing new to purge. It's like `rm something; rm something` then the second invocation is non-zero exit code.

tests/step-103.sh
```diff
@@ -0,0 +1,71 @@
...
+
+echo "Commiting"
+# Make a commit
+echo A >A || exit 1
+git add A || exit 1
+OUTPUT=$(git commit -a -m "Test" 2>&1)
+
+# shellcheck disable=SC2181
+if [ $? -eq 0 ] || ! echo "$OUTPUT" | grep -q "Failed to execute shared hook"; then
```
rycus86: The commit should be successful, and that's how it should be.
This may produce some extra message about non-existing shared hooks which could be suppressed, but I don't think the commit should fail because of it.

tests/step-103.sh
```diff
@@ -0,0 +1,71 @@
...
+# Change url
+(cd ~/.githooks/shared/shared_hooks_102* &&
+    git remote rm origin &&
+    git remote add origin /some/other/url.git) || exit 1
+# Make a commit
+echo A >>A || exit 1
+OUTPUT=$(git commit -a -m "Test" 2>&1)
+
+# shellcheck disable=SC2181
+if [ $? -eq 0 ] || ! (echo "$OUTPUT" | grep "Shared root url" | grep -q "not the same"); then
```
rycus86: Same as above: I expected this commit to work, and it did.

tests/step-103.sh
gabyx: Then you remove line 41. The test on master should fail because Line 38(clone) does not install the shared hooks automatically, and any next commits pass through without running the shared hooks. This is really not wanted, we should really make if fail!
Thats mainly the problem. So removing Line 41. Should make it fail after Line 47.

tests/step-103.sh
gabyx: I strongly think: you placed a `.shared` file into your repo. Meaning you want hooks to be applied. When you clone, the shared hooks are not there. You need manual `git hooks shared update`. Which can get forgotten. 
I made it such that the commit fails, because shared hooks are not present...

gabyx: The problem with these optional behavior is, that user forget things, and I think we should not give the user to much space to really mess up things... git is hard enough :-)...

tests/step-103.sh
gabyx: Ok. Maybe we should make a fail flag in the settings, that any `fail-on-non-existing-shared-hook`
Thats the behavior I would enforce always :-) But maybe users dont want that??...

tests/step-103.sh
rycus86: OK, so as far as I know, ~~there are no hooks run after a `git clone`~~. Oh, I see now that [post-checkout](https://git-scm.com/docs/githooks#_post_checkout) is run after `clone` normally as you found.

Currently, shared hooks are only refreshed at `post-merge` (i.e. when you `git pull` for example), or if you trigger explicitly through the cli tool. I had it like this originally, so that the shared repos are not updated over and over again slowly, when you're trying to do something quickly. Moving between branches with `git checkout` can be expected to complete quickly for example, so I wouldn't do a `git pull` on all the shared hook repos at that point. However, the docs say:
> It is also run after git-clone[1], unless the --no-checkout (-n) option is used. The first parameter given to the hook is the null-ref, the second the ref of the new HEAD and the flag is always 1. Likewise for git worktree add unless --no-checkout is used.

So if we can tell the `git clone` `post-checkout` events apart from the `git checkout` triggered ones, we could refresh shared hooks there as well.

tests/step-103.sh
rycus86: I'd err on the side of being able to complete the task you want (you want to commit & push a change) even if there's a hiccup with one of the hooks. CI should catch those problems too, right? :)

gabyx: Ok. Continuing is the default behavior. But a flag would help in my setup.
- githooks.failOnNonExistingHooks = false
CI can not check everything, for example `default lfs hooks` which are reference in `.shared`
If you dont run these, you have a real problem. So I would really set `githooks.failOnNonExistingHooks = true`

> So if we can tell the git clone post-checkout events apart from the git checkout triggered ones, we could refresh shared hooks there as well.
Jeah, that would be good.
So we check for a null reference if its the `post-checkout` and only then update. Thats sounds good to me.

gabyx:
See the changes https://github.com/rycus86/githooks/pull/22/commits/e020127fc0c3e8d10782b34b3520e4d6859e0313

base-template.sh
```diff
@@ -378,29 +395,30 @@ process_shared_hooks() {
...
+                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Update failed, git pull output:"
                     echo "$PULL_OUTPUT"
                 fi
             else
                 echo "* Retrieving shared hooks from: $SHARED_REPO"
-                CLONE_OUTPUT=$(cd ~/.githooks/shared && git clone "$SHARED_REPO" "$NORMALIZED_NAME" 2>&1)
+                [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
```
rycus86: Out of curiosity, why delete the folder first? Would `git clone` fail if the directory exists?

base-template.sh
```diff
@@ -65,6 +65,9 @@ set_main_variables() {
     if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
         CURRENT_GIT_DIR=".git" # reset to a sensible default
     fi
+
+    # Git null ref
+    GIT_NULL_REF="0000000000000000000000000000000000000000"
```
rycus86: This is only used in `update_shared_hooks_if_appropriate` so I'd declare it there.

base-template.sh
```diff
@@ -378,29 +395,30 @@ process_shared_hooks() {
 #####################################################
 update_shared_hooks_if_appropriate() {
     # run an init/update if we are after a "git pull" or triggered manually
-    if [ "$HOOK_NAME" = "post-merge" ] || [ "$HOOK_NAME" = ".githooks.shared.trigger" ]; then
+    if [ "$HOOK_NAME" = "post-merge" ] ||
```
rycus86: I don't really trust `A || B && C || D` in POSIX shell. Can you please extract them?
Maybe something like:
```
RUN_UPDATE="false"
[ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
[ "$HOOK_NAME" = ".githooks.shared.trigger" ] && RUN_UPDATE="true"
[ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
if [ "$RUN_UPDATE" = "true" ]; then
...
```

base-template.sh
```diff
@@ -421,19 +439,52 @@ update_shared_hooks_if_appropriate() {
...
-        REMOTE_URL=$(cd "$SHARED_ROOT" && git config --get remote.origin.url)
+    # split on comma and newline
+    IFS=",
+    "
+    for SHARED_REPO in $SHARED_REPOS_LIST; do
+
+        set_shared_root "$SHARED_REPO"
+
+        # Fail if the shared root is not available (if enabled)
+        FAIL_ON_NOT_EXISTING=$(git config --global --get githooks.failOnNotExistingSharedHooks)
```
rycus86: To follow the lowercase dot-separated pattern (like `githooks.trust.all`) can you please rename this?
Something like `githooks.fail.on.missing.shared.hooks` perhaps?

base-template.sh
```diff
@@ -421,19 +439,52 @@ update_shared_hooks_if_appropriate() {
...
+        
+        if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+            if [ ! -d "$SHARED_ROOT" ] || [ ! -d "$SHARED_ROOT/.git" ]; then
```
rycus86: `if [ ! -d "$SHARED_ROOT/.git" ]; then` should be enough because if the `.git` directory is not there, it's parent won't be either.

base-template.sh
```diff
@@ -421,19 +439,52 @@ update_shared_hooks_if_appropriate() {
...
+                echo "!   \$ git hooks shared update"
+                return 1
+            fi
+        fi
+
+        # Note: GIT_DIR might be set (?bug?) (actually the case for post-checkout hook)
+        # which means we really need a `-f` to sepcify the actual config!
+        REMOTE_URL=$(cd "$SHARED_ROOT" && git config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
         ACTIVE_REPO=$(echo "$SHARED_REPOS_LIST" | grep -o "$REMOTE_URL")
         if [ "$ACTIVE_REPO" != "$REMOTE_URL" ]; then
```
rycus86: This here is meant to filter out repositories that are in the shared hooks root folder but we don't have them on the `$SHARED_REPOS_LIST` list anymore. With your change, is this still needed?

cli.sh
```diff
@@ -1030,12 +1030,11 @@ list_shared_hook_repos() {
...
-                NORMALIZED_NAME=$(echo "$LIST_ITEM" |
-                    sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                    sed -E "s/[^a-zA-Z0-9]/_/g")
 
-                if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
-                    if [ "$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
+                set_shared_root "$LIST_ITEM"
+                echo "$SHARED_ROOT"
+                if [ -d "$SHARED_ROOT/.git" ]; then
+                    if [ "$(cd "$SHARED_ROOT/.git" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
```
rycus86: Slight change from the old code, we used to `cd` into repo root, not the `repo/.git` folder. Maybe it doesn't make a difference, but if it works, would be nice to go back to that one please.

cli.sh
```diff
@@ -1030,12 +1030,11 @@ list_shared_hook_repos() {
...
         else
             for LIST_ITEM in $(git config --global --get githooks.shared); do
-                NORMALIZED_NAME=$(echo "$LIST_ITEM" |
-                    sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                    sed -E "s/[^a-zA-Z0-9]/_/g")
 
-                if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
-                    if [ "$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
+                set_shared_root "$LIST_ITEM"
+                echo "$SHARED_ROOT"
```
rycus86: Leftover debug statement?

cli.sh
```diff
@@ -1066,12 +1065,11 @@ list_shared_hook_repos() {
...
-                NORMALIZED_NAME=$(echo "$LIST_ITEM" |
-                    sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-                    sed -E "s/[^a-zA-Z0-9]/_/g")
 
-                if [ -d ~/.githooks/shared/"$NORMALIZED_NAME"/.git ]; then
-                    if [ "$(cd ~/.githooks/shared/"$NORMALIZED_NAME" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
+                set_shared_root "$LIST_ITEM"
+
+                if [ -d "$SHARED_ROOT/.git" ]; then
+                    if [ "$(cd "$SHARED_ROOT/.git" && git config --get remote.origin.url)" = "$LIST_ITEM" ]; then
```
rycus86: Same as above, `cd` into the repo root instead of `.git` please if that works.

uninstall.sh
```diff
@@ -162,6 +162,16 @@ uninstall_hooks_from_repo() {
...
 }
 
+############################################################
+# Uninstall shared hooks.
+#
+# Returns:
+#   None
+############################################################
+uninstall_shared_hooks() {
+    rm -rf ~/.githooks/shared >/dev/null 2>&1
```
rycus86: Could follow the same pattern as elsewhere:
```
if [ -d ~/.githooks/shared ]; then
    if ! rm -rf ~/.githooks/shared >/dev/null 2>&1; then
        echo "Failed to delete shared hook repository folders"
        exit 1
    fi
fi
```

uninstall.sh
```diff
@@ -181,12 +191,16 @@ if ! uninstall_from_existing_repositories; then
...
+# Uninstall all shared hooks
+uninstall_shared_hooks
+
 # Unset global Githooks variables
 git config --global --unset githooks.shared
 git config --global --unset githooks.autoupdate.enabled
 git config --global --unset githooks.autoupdate.lastrun
 git config --global --unset githooks.previous.searchdir
 git config --global --unset githooks.disable
+git config --global --unset alias.hooks
```
rycus86: Technically, we could still use the cli with the alias to reinstall, though I think it was probably an oversight from me. If we unset the alias now, we should also delete the cli script.

gabyx: shouldnt we just completely delete the `~/.githooks` folder, without the potential `template` subfolder.. or also with it? because it anyways contains githooks templates and not own ones... hm...

cli.sh
gabyx: ah jeah!

cli.sh
gabyx: good catch, thx

base-template.sh
gabyx: not really, but if somebody messes around with the repo in the shared root folder, this asserts that the url in the root is consistent with what we want from the `.shared`. I would leave it in... ?

base-template.sh
gabyx: That would actuall insert
```ini
[githooks "fail.on.missing.shared"]
	hooks = true
``` 
I dont think you want that :-)

gabyx: May be, `githooks.fail.onNotExistingSharedHooks` would work.

base-template.sh
gabyx: jep, git clone would fail, saying already existing...

uninstall.sh
gabyx: No I think its not good, since the template folder is choosen by the user, we leave it there, but delete all other stuff `bin,shared` folders...

gabyx: Unexcepted behavior if you have user defined hooks in `.githooks/templates` which then get correctly back replaced but at the end deleted, which is stupid...

base-template.sh
rycus86: OK, let's leave it as is.

base-template.sh
```diff
@@ -421,19 +441,52 @@ update_shared_hooks_if_appropriate() {
...
-        REMOTE_URL=$(cd "$SHARED_ROOT" && git config --get remote.origin.url)
+    # split on comma and newline
+    IFS=",
+    "
+    for SHARED_REPO in $SHARED_REPOS_LIST; do
+
+        set_shared_root "$SHARED_REPO"
+
+        # Fail if the shared root is not available (if enabled)
+        FAIL_ON_NOT_EXISTING=$(git config --global --get githooks.failOnNotExistingSharedHooks)
```
rycus86: I'd drop the `--global` so you can either set this globally or by repository.

base-template.sh
```diff
@@ -421,19 +441,52 @@ update_shared_hooks_if_appropriate() {
...
+
+        if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+            if [ ! -d "$SHARED_ROOT/.git" ]; then
```
rycus86: Maybe test `[ ! -f "$SHARED_ROOT/.git/config" ]` because that is what we're pointing at next.

base-template.sh
```diff
@@ -421,19 +441,52 @@ update_shared_hooks_if_appropriate() {
...
+                echo "! Failed to execute shared hook: $SHARED_REPO"
```
rycus86: Please update to:
```
echo "! Failed to execute shared hooks in $SHARED_REPO"
echo "! It is not available. To fix, run:"
echo "!   \$ git hooks shared update"
return 1
```

base-template.sh
```diff
@@ -421,19 +441,52 @@ update_shared_hooks_if_appropriate() {
...
+            fi
+        fi
+
+        # Note: GIT_DIR might be set (?bug?) (actually the case for post-checkout hook)
+        # which means we really need a `-f` to sepcify the actual config!
+        REMOTE_URL=$(cd "$SHARED_ROOT" && git config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
         ACTIVE_REPO=$(echo "$SHARED_REPOS_LIST" | grep -o "$REMOTE_URL")
         if [ "$ACTIVE_REPO" != "$REMOTE_URL" ]; then
-            continue
+            echo "! Failed to execute shared hook: $SHARED_REPO"
```
rycus86: Please update to:
```
echo "! Failed to execute shared hooks in $SHARED_REPO"
echo "! The URL \`$REMOTE_URL\` is different."
echo "To fix it, run:"
echo "!   \$ git hooks shared purge"
echo "!   \$ git hooks shared update"
```

base-template.sh
```diff
@@ -421,19 +441,52 @@ update_shared_hooks_if_appropriate() {
...
+            echo "! Shared root url \`$REMOTE_URL\` is not the same as "
+            echo "! the requested url \`$SHARED_REPO\`. Do run:"
+            echo "!   \$ git hooks shared purge"
+            echo "!   \$ git hooks shared update"
+
+            if [ "$FAIL_ON_NOT_EXISTING" = "true"]; then
+                return 1
+            else
+                echo "!   Continue..."
```
rycus86: Please update to:
```
echo "!   Continuing..."
```

tests/step-103.sh
```diff
@@ -0,0 +1,71 @@
...
+# Remove all shared hooks
+git hooks shared purge || exit 1
+
+# Clone a new one
+echo "Cloning"
+cd /tmp || exit 1
+git clone /tmp/test103 test103-clone && cd test103-clone || exit 1
+
+# Remove all shared hooks
+git hooks shared purge || exit 1
```
rycus86: Can you please also add a test that fails on the double purge with the new config option set to true?

gabyx:
Thx will look at it this eveninh maybe

Von meinem iPhone gesendet

gabyx:
Finally all  tests good :-)

base-template.sh
```diff
@@ -378,29 +392,35 @@ process_shared_hooks() {
...
 update_shared_hooks_if_appropriate() {
     # run an init/update if we are after a "git pull" or triggered manually
-    if [ "$HOOK_NAME" = "post-merge" ] || [ "$HOOK_NAME" = ".githooks.shared.trigger" ]; then
+    GIT_NULL_REF="0000000000000000000000000000000000000000"
+
+    RUN_UPDATE="false"
+    [ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
+    [ "$HOOK_NAME" = ".githooks.shared.trigger" ] && RUN_UPDATE="true"
+    [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
+    echo "Hook name: $HOOK_NAME, $1"
```
rycus86: Is this a debugging statement maybe that was left in?

gabyx: Ou jeah, damm!

gabyx: Thx for beeing, harsh :-)

rycus86: Hehe, no, thanks for fixing it up!
I'll try to get it merged in an hour or so. 🤞

rycus86:
Thanks for this again @gabyx !

PR #23, opened 2019-07-13
Dispatch User Prompts to Dialog Tool

gabyx:
Introduce a small tool "dialog" which dispatches all dialog prompts in `base-template.sh` to a script.
An example implementation with python3.6 and higher is given in `examples/tools/dialog`.

### DIFF
--- a/examples/tools/dialog/run
+++ b/examples/tools/dialog/run
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+
+import sys
+import platform
+import subprocess
+from tkinter import Tk
+from tkinter import dialog
+
+
+def ask(title, text, strings=('Yes', 'No'), bitmap='warning', default=0):
+    """ Making a dialog windows.
+
+    Args:
+        title (str): Title of the windows
+        text (str): Text to be show to the user
+        strings (tuple, optional): Button strings to be shown. Defaults to ('Yes', 'No').
+        bitmap (str, optional): Bitmap to use for the icon. Defaults to 'warning'.
+        default (int, optional): Index of the default button. Defaults to 0.  
+    Returns:
+        [type]: [description]
+    """
+
+    d = dialog.Dialog(title=title,
+                      text=text,
+                      bitmap=bitmap,
+                      default=default,
+                      strings=strings)
+    return d.num
+
+
+if __name__ == '__main__':
+    title = sys.argv[1]
+    text = sys.argv[2]
+    options = sys.argv[3].strip().split("/")
+
+    longOptions = []
+    if len(sys.argv) >= 5:
+        longOptions = [l.strip() for l in sys.argv[4:]]
+
+    # Fill up the long options if not given
+    if len(longOptions) < len(options):
+        longOptions += options[len(longOptions):]
+
+    default = 0
+    for idx, o in enumerate(options):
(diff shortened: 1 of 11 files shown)


gabyx:
Works nicely:
image

examples/tools/dialog/run
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
```
rycus86: Why is this needed? Can't we just have the Python script with executable flags? The `/usr/bin/env python` bit should take care of the interpreter, and there's nothing in there that needs `PYTHONPATH` to be set to anything special?

rycus86:
I'm a bit concerned about trying to detect if we're running from a connected terminal in a cross-platform way. Can't we just check if we have a dialog tool configured, and if we do use that otherwise just try reading from the terminal as before?

examples/tools/dialog/run
gabyx: This was a copy of my setup/. On windows its not so easy. On Linux for sure. Lets do it the easy way.

gabyx:
> I'm a bit concerned about trying to detect if we're running from a connected terminal in a cross-platform way. Can't we just check if we have a dialog tool configured, and if we do use that otherwise just try reading from the terminal as before?

I think it should work, shouldn it?. The thing, is the dialog tool should really only be the case when running without a terminal attached. Because when running in the terminal you can just answer there instead of the GUI. Or should we do it like you said. and I can provide for my solution code in the `run` script to detect if we have a terminal attached. 
So far the provided solution works. At least on Windows with Git-Bash (from the official git install) etc.
Should also work in Linux.

gabyx:
At least I tested it on MacOs as well.

rycus86:
OK, I think this looks good to me, I'd just like to start with keeping everything as-is, except if you configure a dialog tool, in which case that is invoked instead of the `read -r`. Could you change it like that and squash please? Thanks!

gabyx:
Ok, cool!

gabyx:
Force pushed.

rycus86:
This looks good, just needs a rebase/merge for the conflicts and a squash after perhaps. There's a few typos here and there, but I can take care of them after merging this PR.

gabyx:
Maybe we could also show the stderr of the runned hook over the dialog tool, maybe, maybe in another PR :-)

rycus86:
Sorry, had to revert this one, was breaking all the user prompts without the dialog tool. :(
I think the problem was here:
```
    else
        # Read from stdin (because its connected)
        # shellcheck disable=SC2059
        printf "$TEXT $HINT_TEXT [$SHORT_OPTIONS]: "
        read -r VARIABLE
    fi
```

This looks like we were loading user input into `$VARIABLE` instead of the actual variable name. I *think* `read -r $VARIABLE` could work though.
Would you like to reopen this and give it another try?
I think where this slipped through, is that we replaced the whole call to `show_prompt` in the test setup phase with the default answer. Instead of this, we should perhaps just change `call_script "$DIALOG_TOOL"` to return what we want it to return.

rycus86:
@gabyx :arrow_up: in case you haven't seen

gabyx:
Oh, I think you are right.

gabyx:
I reopen this, :-)

gabyx:
See #28

PR #26, opened 2019-07-17
Bugfix/disturbing output current git dir

gabyx:
Output:
```shell
fatal: not a git repository (or any of the parent directories): .git
```
on `git hooks tools` when not in a git repo.
I also dont understand the motivation behing: 
https://github.com/rycus86/githooks/blob/dbdb37bcb790d72a2d7ae262087508d8f2e6fe35/cli.sh#L66

Thanks for correcting this. :-)


rycus86:
That's apparently the output when `git rev-parse --git-common-dir` is not pointing to an override or something? 🤷‍♂️

Thanks for fixing it up!

PR #28, opened 2019-07-22
Dispatch user prompts to dialog tool. :heart:

gabyx:
Another try. Fixed the `read -r VARIABLE`

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1907.221206-176cca
+# Version: 1907.261441-38baa8
 
 #####################################################
 # Execute the current hook,
@@ -256,8 +256,7 @@ is_trusted_repo() {
         # shellcheck disable=SC2181
         if [ $TRUST_ALL_RESULT -ne 0 ]; then
             echo "! This repository wants you to trust all current and future hooks without prompting"
-            printf "  Do you want to allow running every current and future hooks? [y/N] "
-            read -r TRUST_ALL_HOOKS </dev/tty
+            show_prompt TRUST_ALL_HOOKS "  Do you want to allow running every current and future hooks?" "" "y/N" "Yes" "No"
 
             if [ "$TRUST_ALL_HOOKS" = "y" ] || [ "$TRUST_ALL_HOOKS" = "Y" ]; then
                 git config githooks.trust.all Y
@@ -308,8 +307,7 @@ execute_opt_in_checks() {
         if [ "$ACCEPT_CHANGES" = "a" ] || [ "$ACCEPT_CHANGES" = "A" ]; then
             echo "  Already accepted"
         else
-            printf "  Do you you accept the changes? (Yes, all, no, disable) [Y/a/n/d] "
-            read -r ACCEPT_CHANGES </dev/tty
+            show_prompt ACCEPT_CHANGES "  Do you accept the changes?" "(Yes, all, no, disable)" "Y/a/n/d" "Yes" "All" "No" "Disable"
 
             if [ "$ACCEPT_CHANGES" = "n" ] || [ "$ACCEPT_CHANGES" = "N" ]; then
                 echo "* Not running $HOOK_FILE"
@@ -593,6 +591,57 @@ call_script() {
 }
 
 #####################################################
+# Show a prompt with the text `$2` with
+#   hint text `$3` with
+#   options `$4` in form of e.g. `Y/a/n/d` and
+#   optional long options [optional]:
+#   e.g.
+#    `$5-$8` : "Yes" "All" "None" "Disable"
+#   First capital short option character is treated
+#   as default.
+#
+#####################################################
+show_prompt() {
+    DIALOG_TOOL="$(get_tool_script "dialog")"
+    VARIABLE="$1"
+    shift
+    TEXT="$1"
+    shift
+    HINT_TEXT="$1"
+    shift
+    SHORT_OPTIONS="$1"
+    shift
+
+    if [ "$DIALOG_TOOL" != "" ]; then
+        ANSWER=$(call_script "$DIALOG_TOOL" "githook::" "$TEXT" "$SHORT_OPTIONS" "$@")
+        # shellcheck disable=SC2181
+        if [ $? -eq 0 ]; then
+
+            if ! echo "$SHORT_OPTIONS" | grep -q "$ANSWER"; then
+                echo "! Dialog tool did return wrong answer $ANSWER -> Abort."
+                exit 1
+            fi
+
+            # Safeguard `eval`
+            if ! echo "$VARIABLE" | grep -qE "^[A-Z_]+\$"; then
+                echo "! Invalid variable name: $VARIABLE"
+                exit 1
+            fi
+
+            eval "$VARIABLE"="$ANSWER"
+            return
+        fi
+        # Running fallback...
+    fi
+
+    # Read from stdin (because its connected)
+    printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
+    # shellcheck disable=SC2229
+    read -r "$VARIABLE"
+
+}
+
+#####################################################
 # Downloads a file "$1" with `wget` or `curl`
 #
 # Returns:
@@ -712,8 +761,7 @@ is_single_repo() {
 #####################################################
 should_run_update() {
     echo "* There is a new Githooks update available: Version $LATEST_VERSION"
-    printf "    Would you like to install it now? [Y/n] "
-    read -r EXECUTE_UPDATE </dev/tty
+    show_prompt EXECUTE_UPDATE "    Would you like to install it now?" "" "Y/n" "Yes" "no"
 
     if [ -z "$EXECUTE_UPDATE" ] || [ "$EXECUTE_UPDATE" = "y" ] || [ "$EXECUTE_UPDATE" = "Y" ]; then
         return 0
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1907.221206-176cca
+# Version: 1907.261441-38baa8
 
 #####################################################
 # Prints the command line help for usage and
@@ -1969,12 +1969,14 @@ manage_tools() {
     if [ "$1" = "help" ]; then
         print_help_header
         echo "
-git hooks tools register [download] <scriptFolder>
+git hooks tools register [download|dialog] <scriptFolder>
 
     ( experimental feature )
 
     Install the script folder \`<scriptFolder>\` in 
     the installation directory under \`tools/<toolName>\`.
+
+    Download Tool:
     The interface of the tool is as follows.
     
     # if \`run\` is executable
@@ -1982,11 +1984,30 @@ git hooks tools register [download] <scriptFolder>
     # otherwise, assuming \`run\` is a shell script
     \$ sh run <relativeFilePath> <outputFile>
     
-    The arguments for the download tool are:
+    The arguments are:
     - \`<relativeFilePath>\` is the file relative to the repository root
     - \`<outputFile>\` file to write the results to (may not exist yet)
 
-git hooks tools unregister [download]
+    Dialog Tool:
+    The interface of the tool is as follows.
+    
+    # if \`run\` is executable
+    \$ run <title> <text> <options> <long-options>
+    # otherwise, assuming \`run\` is a shell script
+    \$ sh run <title> <text> <options> <long-options>
+
+    The arguments for are:
+    - \`<title>\` the title for the user gui dialog
+    - \`<text>\` the text for the user gui dialog
+    - \`<short-options>\` the button return values, slash-delimited, 
+        e.g. \`Y/n/d\`.
+        The default button is the first capital character found.
+    - \`<long-options>\` the button texts in the gui,
+        e.g. \`Yes/no/disable\`
+
+    The script needs to return one of the short-options on \`stdout\`.
+
+git hooks tools unregister [download|dialog]
 
     ( experimental feature )
 
@@ -2022,7 +2043,7 @@ git hooks tools unregister [download]
 #   1 on failure, 0 otherwise
 #####################################################
 tools_install() {
-    if [ "$1" = "download" ]; then
+    if [ "$1" = "download" ] || [ "$1" = "dialog" ]; then
         SCRIPT_FOLDER="$2"
 
         if [ -d "$SCRIPT_FOLDER" ]; then
@@ -2051,7 +2072,7 @@ tools_install() {
             exit 1
         fi
     else
-        echo "! Invalid operation: \`$1\` (use \`download\`)"
+        echo "! Invalid operation: \`$1\` (use \`download\` or  \`dialog\`)"
         exit 1
     fi
 }
@@ -2065,7 +2086,7 @@ tools_install() {
 tools_uninstall() {
     [ "$2" = "--quiet" ] && QUIET="Y"
 
-    if [ "$1" = "download" ]; then
+    if [ "$1" = "download" ] || [ "$1" = "dialog" ]; then
         if [ -d ~/".githooks/tools/$1" ]; then
             rm -r ~/".githooks/tools/$1"
             [ -n "$QUIET" ] || echo "Uninstalled the \`$1\` tool"
@@ -2073,7 +2094,7 @@ tools_uninstall() {
             [ -n "$QUIET" ] || echo "! The \`$1\` tool is not installed"
         fi
     else
-        [ -n "$QUIET" ] || echo "! Invalid tool: \`$1\` (use \`download\`)"
+        [ -n "$QUIET" ] || echo "! Invalid tool: \`$1\` (use \`download\` or \`dialog\`)"
         exit 1
     fi
 }
--- a/tests/coverage.sh
+++ b/tests/coverage.sh
@@ -38,9 +38,8 @@ RUN find /var/lib -name '*.sh' -exec sed -i 's|#!/bin/sh|#!/bin/bash|g' {} \\; &
 # Change the base template so we can pass in the hook name and accept flags
     sed -i 's|HOOK_NAME=.*|HOOK_NAME=\${HOOK_NAME:-\$(basename "\$0")}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|HOOK_FOLDER=.*|HOOK_FOLDER=\${HOOK_FOLDER:-\$(dirname "\$0")}|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|ACCEPT_CHANGES=.*|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|read -r ACCEPT_CHANGES|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests: read -r ACCEPT_CHANGES|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|read -r TRUST_ALL_HOOKS|TRUST_ALL_HOOKS=\${TRUST_ALL_HOOKS} # disabled for tests: read -r TRUST_ALL_HOOKS|' /var/lib/githooks/base-template.sh
+    sed -i 's|ACCEPT_CHANGES=|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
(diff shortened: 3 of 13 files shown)


gabyx:
I think there is a missing:
```
    sed -i 's|show_prompt TRUST_ALL_HOOKS.*|TRUST_ALL_HOOKS=\${TRUST_ALL_HOOKS} # disabled for tests: read -r TRUST_ALL_HOOKS|' /var/lib/githooks/base-template.sh

``` 
in `exec-tests.sh`... 
Test 34 is failing, strange? Do you know why...

gabyx:
Is this test really correct??
https://github.com/rycus86/githooks/blob/df9e879fd9c834f6c23b7138e71025e69d8de235/tests/step-034.sh#L12

https://github.com/rycus86/githooks/blob/df9e879fd9c834f6c23b7138e71025e69d8de235/tests/step-034.sh#L22
You say here `dont accept changes` and it does not run the hooks..??

base-template.sh
```diff
@@ -592,6 +590,48 @@ call_script() {
...
+        fi
+        if ! echo "$SHORT_OPTIONS" | grep -q "$ANSWER"; then
+            echo "! Dialog tool did return wrong answer $ANSWER -> Abort."
+            exit 1
+        fi
+        eval "$VARIABLE"="$ANSWER"
+    else
+        # Read from stdin (because its connected)
+        # shellcheck disable=SC2059
+        printf "$TEXT $HINT_TEXT [$SHORT_OPTIONS]: "
```
rycus86: Instead of disabling this check, could you change it to `printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"` ?

base-template.sh
```diff
@@ -592,6 +590,48 @@ call_script() {
...
+        # shellcheck disable=SC2181
+        if [ $? -ne 0 ]; then
+            echo "! Dialog tool did not succeed -> Abort."
+            exit 1
+        fi
+        if ! echo "$SHORT_OPTIONS" | grep -q "$ANSWER"; then
+            echo "! Dialog tool did return wrong answer $ANSWER -> Abort."
+            exit 1
+        fi
+        eval "$VARIABLE"="$ANSWER"
```
rycus86: `eval` makes me a bit nervous but haven't found anything POSIX as an alternative. Could we at least validate it's nothing dangerous, like only uppercase characters with underscores?
I think this should do the trick:
```
if ! echo "$VARIABLE" | grep -qE "^[A-Z_]+\$"; then
    echo "! Invalid variable name: $VARIABLE"
    exit 1
fi
```

cli.sh
```diff
@@ -2065,15 +2086,15 @@ tools_install() {
...
+    if [ "$1" = "download" ] || [ "$1" = "dialog" ]; then
         if [ -d ~/".githooks/tools/$1" ]; then
             rm -r ~/".githooks/tools/$1"
             [ -n "$QUIET" ] || echo "Uninstalled the \`$1\` tool"
         else
             [ -n "$QUIET" ] || echo "! The \`$1\` tool is not installed"
         fi
     else
-        [ -n "$QUIET" ] || echo "! Invalid tool: \`$1\` (use \`download\`)"
+        [ -n "$QUIET" ] || echo "! Invalid tool: \`$1\` (use \`download\` or \`dailog\`)"
```
rycus86: `dailog` -> `dialog`

tests/coverage.sh
```diff
@@ -39,8 +39,9 @@ RUN find /var/lib -name '*.sh' -exec sed -i 's|#!/bin/sh|#!/bin/bash|g' {} \\; &
     sed -i 's|HOOK_NAME=.*|HOOK_NAME=\${HOOK_NAME:-\$(basename "\$0")}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|HOOK_FOLDER=.*|HOOK_FOLDER=\${HOOK_FOLDER:-\$(dirname "\$0")}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|ACCEPT_CHANGES=.*|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|read -r ACCEPT_CHANGES|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests: read -r ACCEPT_CHANGES|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|read -r TRUST_ALL_HOOKS|TRUST_ALL_HOOKS=\${TRUST_ALL_HOOKS} # disabled for tests: read -r TRUST_ALL_HOOKS|' /var/lib/githooks/base-template.sh
+    sed -i 's|show_prompt ACCEPT_CHANGES.*|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests|' /var/lib/githooks/base-template.sh \\
```
rycus86: This is what got us in trouble last time. It mocks the whole prompt logic out, rather than the variable reading bit. I think the change should be (if it works, haven't tested):
```
sed -i 's|read -r "\$VARIABLE"|eval "\$VARIABLE"="\$TEST_ANSWER" # disabled for tests: read -r "\$VARIABLE"|' /var/lib/githooks/base-template.sh && \\
sed -i 's|(call_script "\$DIALOG_TOOL" "githook::" "\$TEXT" "\$SHORT_OPTIONS" "\$@")|(echo "\$TEST_ANSWER") # disabled for tests: call_script "\$DIALOG_TOOL" "githook::" "\$TEXT" "\$SHORT_OPTIONS" "\$@"|' /var/lib/githooks/base-template.sh
```

tests/exec-tests.sh
```diff
@@ -14,11 +14,13 @@ ADD tests/exec-steps.sh tests/step-* /var/lib/tests/
 
 # Do not use the terminal in tests
 RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
+    sed -i 's|show_prompt ACCEPT_CHANGES.*|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests|' /var/lib/githooks/install.sh && \\
```
rycus86: Is this needed? We don't show (GUI) prompts during installation.

tests/exec-tests.sh
```diff
@@ -14,11 +14,13 @@ ADD tests/exec-steps.sh tests/step-* /var/lib/tests/
...
 # Change the base template so we can pass in the hook name and accept flags
     sed -i 's|HOOK_NAME=.*|HOOK_NAME=\${HOOK_NAME:-\$(basename "\$0")}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|HOOK_FOLDER=.*|HOOK_FOLDER=\${HOOK_FOLDER:-\$(dirname "\$0")}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|ACCEPT_CHANGES=.*|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
-    sed -i 's|read -r ACCEPT_CHANGES|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests: read -r ACCEPT_CHANGES|' /var/lib/githooks/base-template.sh
+    sed -i 's|show_prompt ACCEPT_CHANGES.*|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests|' /var/lib/githooks/base-template.sh \\
```
rycus86: Similar change as for the coverage I think.

tests/step-032.sh
```diff
@@ -22,7 +22,7 @@ sed -i 's/^# Version: .*/# Version: 0/' /var/lib/githooks/base-template.sh &&
     exit 1
 
 OUTPUT=$(
-    sed -i 's|read -r EXECUTE_UPDATE </dev/tty|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
+    sed -i 's|show_prompt EXECUTE_UPDATE.*|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
```
rycus86: `sed -E` is needed here I think.

tests/step-042.sh
```diff
@@ -51,7 +51,7 @@ sed -i 's/^# Version: .*/# Version: 0/' /var/lib/githooks/base-template.sh ||
     exit 1
 
 OUTPUT=$(
-    sed -i 's|read -r EXECUTE_UPDATE </dev/tty|EXECUTE_UPDATE="Y"|' /var/lib/githooks/base-template.sh &&
+    sed -i 's|show_prompt EXECUTE_UPDATE.*|EXECUTE_UPDATE="Y"|' /var/lib/githooks/base-template.sh &&
```
rycus86: `sed -E`

tests/step-076.sh
```diff
@@ -22,7 +22,7 @@ sed -i 's/^# Version: .*/# Version: 0/' /var/lib/githooks/base-template.sh &&
     exit 1
 
 OUTPUT=$(
-    sed -i 's|read -r EXECUTE_UPDATE </dev/tty|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
+    sed -i 's|show_prompt EXECUTE_UPDATE.*|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
```
rycus86: `sed -E`

tests/step-077.sh
```diff
@@ -23,7 +23,7 @@ sed -i 's/^# Version: .*/# Version: 0/' /var/lib/githooks/base-template.sh &&
     exit 1
 
 OUTPUT=$(
-    sed -i 's|read -r EXECUTE_UPDATE </dev/tty|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
+    sed -i 's|show_prompt EXECUTE_UPDATE.*|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
```
rycus86: `sed -E`

base-template.sh
gabyx: Jeah, makes me nervous as well, I couldnt find something better as well unfortunately.
Lets do this, jup!

tests/coverage.sh
gabyx: hm... but with your solution we would inject `TEST_ANSWER="asdasd"`. 
But my code just skips the prompt for the **corresponding variable** an replaces it, shouldn't that be fine, its the exact same line where the former `read -r` was...:
```bash
    sed -iE 's|show_prompt ACCEPT_CHANGES.*|echo "Accepted: \$ACCEPT_CHANGES" # disabled for tests|' /var/lib/githooks/base-template.sh \\
    sed -iE 's|show_prompt TRUST_ALL_HOOKS.*|TRUST_ALL_HOOKS=\${TRUST_ALL_HOOKS} # disabled for tests|' /var/lib/githooks/base-template.sh
```

tests/exec-tests.sh
gabyx: Needed, because it will install the base-template, where we also need to mock the prompts, right?

tests/coverage.sh
rycus86: The problem here was that the new `read ...` was actually broken, but the test code necer reached there.

tests/exec-tests.sh
rycus86: The install script doesn't have the new `show_prompt` function, so it can't call it, right?

tests/coverage.sh
gabyx: Aha, true, because of the missing `"$VARIABLE"`.
I think I leave the replacement as I did it, because with your solution, its kind of hard to tell which prompt you mean by injecting the mock `TEST_ANSWER="Y"`... lets see if the tests work...

tests/exec-tests.sh
gabyx: Hm.. thats true. But when we use the install script `install.sh` in the tests the base_template for example in a `git clone` afterwards has all `show_prompts` in it. I dont know exactly if that is good?
Especially for `step-103.sh` we need to mock all prompts, hm... I am quite unsure, how I should do it.
I thought I mock everything out in `install.sh` as well as the used `/var/lib/githookts/base-template.sh` for all tests...??

gabyx: General thought:
Shouldnt we mock every possible `show_prompt` no matter where, and in the tests inject the required variables there: In the base-template we have 3 prompts:
```
show_prompt TRUST_ALL_HOOKS "  Do you want to allow running every current and future hooks?" "" "y/N" "Yes" "No"

show_prompt ACCEPT_CHANGES "  Do you you accept the changes?" "(Yes, all, no, disable)" "Y/a/n/d" "Yes" "All" "No" "Disable"

show_prompt EXECUTE_UPDATE "    Would you like to install it now?" "" "Y/n" "Yes" "no"
```
Shouldn't we replace these 3 occurances globaly, and also in the `install.sh`?

gabyx: **Question**: Why does test-103.sh work. I have not replaced `show_prompt TRUST_ALL_HOOKS` in the `install.sh`. Shouldn't pop up the "trust" prompt ?? Confused, sorry...

gabyx: See the latest push, the tests work now.

tests/coverage.sh
rycus86: OK, that wasn't working very well, so I tried reusing the previous variables set in the tests already.
It was super painful, but this seems to work well:
```
    sed -i 's%ANSWER=\$(call_script "\$DIALOG_TOOL" "githook::" "\$TEXT" "\$SHORT_OPTIONS" "\$@")%eval "ANSWER=\\\\\$\$(eval echo "\\\\\$VARIABLE")" # disabled for tests: ANSWER=\$(call_script "\$DIALOG_TOOL" "githook::" "\$TEXT" "\$SHORT_OPTIONS" "\$@")%' /var/lib/githooks/base-template.sh
```
Can you please use this?
Also, would you mind duplicating one of the test cases and create a file at `~/".githooks/tools/dialog/run"` so the new code path is also tested?

Thanks!

tests/exec-tests.sh
```diff
@@ -14,11 +14,15 @@ ADD tests/exec-steps.sh tests/step-* /var/lib/tests/
 
 # Do not use the terminal in tests
 RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
+    sed -i -E 's|ACCEPT_CHANGES=.*|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/install.sh && \\
```
rycus86: I tested this on your branch, we do not need to change the install script with these 3 lines, please remove them.

tests/coverage.sh
gabyx: Of course I can use, do you might enligthen me, why my part was not working? You rather like to mocke the show_prompt. right? So your code does that, and mine skips it. ok, fair enough :).
Thanks for the painful support.

rycus86: Yeah, your code was working fine. 👌
It just meant that the whole `show_prompt` function was skipped completely, and the bit that was broken on the first try was in there, so I wanted to see if we can test most of it and just mock out getting the answer.

gabyx:
I added a new test 104 and changed to your suggestions.
Question: I found still some caveats, can we make the interface of the dialog tool a little better.
Returnin the answer on stdout is a bit cumbersome, as we cannot  switch inside the dialog if we have terminal for example and then print a prompt and so on...
``` shell
        ANSWER=$(call_script "$DIALOG_TOOL" "githook::" "$TEXT" "$SHORT_OPTIONS" "$@")
```
- we could use the exit code > 0 for the success and answer
- we could use stderr for the answer (ugly)
- we could simply fallback to the `read -r` when the dialog returns exit code 1 (use default read -r), exit code > 1 means failure

Option 3: We can still detect in the dialog `run` if we are interactive or not...

gabyx:
@rycus86 : Could you fix the following issue:
in `coverage.sh` you replace the inline content with the base-template (`cat..`)
```
# Replace the inline content with loading the source file
    python /var/lib/tests/replace-inline-content.py /var/lib/githooks && \\
```
And in this way running the `install.sh` it uses the mocked/sed-replaced content, and test-104 fails.
Is that an error?; Maybe make a backup and use this in `replace-inline-content.py`
I probably dont really get the test/coverage setup, maybe you could enlighten me a bit:
As I see you want 
- the install script not mocked, so everything is installed as we commited it
  (you only replace /dev/tty, to make the install run through... I think)
- the `base-template.sh` mocked for special tests and prompt overrides etc...

gabyx:
Hopefully ready to merge, if you are :-) Thx.

gabyx:
You can squash merge :-) probably

rycus86:
> @rycus86 : Could you fix the following issue:
> in `coverage.sh` you replace the inline content with the base-template (`cat..`)
> 
> ```
> # Replace the inline content with loading the source file
>     python /var/lib/tests/replace-inline-content.py /var/lib/githooks && \\
> ```
> 
> And in this way running the `install.sh` it uses the mocked/sed-replaced content, and test-104 fails.
> Is that an error?; Maybe make a backup and use this in `replace-inline-content.py`
> I probably dont really get the test/coverage setup, maybe you could enlighten me a bit:
> As I see you want
> 
> * the install script not mocked, so everything is installed as we commited it
>   (you only replace /dev/tty, to make the install run through... I think)
> * the `base-template.sh` mocked for special tests and prompt overrides etc...
> 
> **Fixed it in [823a5eb](https://github.com/rycus86/githooks/commit/823a5ebb1a1f64f7a021887a3c9e6d3889695aa9)**

You can see the generated files we run the coverage against if you open the `./cover/index.html` file in a browser then click through to the install script's coverage. The line with the `cat` will turn into this for example:
```
BASE_TEMPLATE_CONTENT=$(cat /var/lib/githooks/base-template.sh) #> BASE_TEMPLATE_CONTENT='#!/bin/bash
```

So this makes sure we use the contents of the base template and the cli script of the current commit.

I checked out your branch and have now worked out why it wasn't working. It's because we mock out the response of the script, without actually calling it. I'm OK with that for now, so please revert the last commit, and in the test, just have `ACCEPT_CHANGES=Y` plus make sure the dialog file exists at the expected location so the new code path is triggered.

base-template.sh
```diff
@@ -592,6 +590,60 @@ call_script() {
...
+
+            # Safeguard `eval`
+            if ! echo "$VARIABLE" | grep -qE "^[A-Z_]+\$"; then
+                echo "! Invalid variable name: $VARIABLE"
+                exit 1
+            fi
+
+            eval "$VARIABLE"="$ANSWER"
+            return
+        elif [ $? -ge 2 ]; then
```
rycus86: Can we just have all failures fall back to terminal please?

tests/step-032.sh
```diff
@@ -22,7 +22,7 @@ sed -i 's/^# Version: .*/# Version: 0/' /var/lib/githooks/base-template.sh &&
     exit 1
 
 OUTPUT=$(
-    sed -i 's|read -r EXECUTE_UPDATE </dev/tty|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
+    sed -i -E 's|show_prompt EXECUTE_UPDATE.*|EXECUTE_UPDATE="N"|' /var/lib/githooks/base-template.sh &&
```
rycus86: This could now be changed to 
```
OUTPUT=$(
    HOOK_NAME=post-commit HOOK_FOLDER=$(pwd)/.git/hooks ACCEPT_CHANGES=A EXECUTE_UPDATE=N \
        sh /var/lib/githooks/base-template.sh
)
```

Can you try it please? If it doesn't work on first try, just leave it, I'll sort it out later.

rycus86:
@gabyx left a few comments but otherwise looking good. :+1: 
Please do those changes and squash the commits when you can, then I'll merge this in after as soon as I have time.
Thanks a lot!

base-template.sh
gabyx: we can  do that. yes.

gabyx:
is the `./cover/index.html` generated in the coverage.sh run?  because I dont see any...

rycus86:
> is the `./cover/index.html` generated in the coverage.sh run? because I dont see any...

Yeah, if you run `sh tests/coverage.sh`, it should generate it in the `cover` directory of the root of the project.

gabyx:
Where would you set ACCEPT_CHANGES=Y?
in Test `step-104.sh` I need to trigger one prompt to trigger the installed dialog tool.
But the install.sh has every prompt mocked...

rycus86:
Just on the commit instruction like `ACCEPT_CHANGES=Y git commit ...`

rycus86:
So the actual dialog tool won't run, because that will be something the user chooses, so we don't need to test that. Maybe we could, I'm not sure.

Thinking about it now, perhaps we should only mock the `read -r` output and let the test dialog tool run if exists?

gabyx:
Ah I need to trigger the code by executing directly `base-template.sh` ?

rycus86:
No, running `git commit` should trigger the pre-commit hook just fine.

gabyx:
Perfect, now it works.

rycus86:
Looks good, thanks a lot! Will merge it in tomorrow. :tada:

rycus86:
Sorry for the delay, just want to fix up some formatting after merge and can't find the time. Will try later today (I hope).

gabyx:
Great :-)

PR #29, opened 2019-07-28
Whitespaces in paths break for loop

gabyx:
I searched for all `find` statements and found some which do not work with 
whitespaces in path names, I corrected them.

Thanks for reviewing this.

### DIFF
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1907.302109-df1e4b
+# Version: 1907.311448-fee6c0
 
 #####################################################
 # Prints the command line help for usage and
@@ -66,6 +66,10 @@ set_main_variables() {
     if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
         CURRENT_GIT_DIR=".git"
     fi
+
+    # Global IFS for loops
+    IFS_COMMA_NEWLINE=",
+"
 }
 
 #####################################################
@@ -225,7 +229,7 @@ git hooks disable [-r|--reset]
 
     ensure_checksum_file_exists
 
-    for HOOK_FILE in $(find "$HOOK_PATH" -type f | grep "/.githooks/"); do
+    find "$HOOK_PATH" -type f -path "*/.githooks/*" | while IFS= read -r HOOK_FILE; do
         if grep -q "disabled> $HOOK_FILE" "${CURRENT_GIT_DIR}/.githooks.checksum" 2>/dev/null; then
             echo "Hook file is already disabled at $HOOK_FILE"
             continue
@@ -312,7 +316,7 @@ git hooks accept [trigger]
     find_hook_path_to_enable_or_disable "$@" || exit 1
     ensure_checksum_file_exists
 
-    for HOOK_FILE in $(find "$HOOK_PATH" -type f | grep "/.githooks/"); do
+    find "$HOOK_PATH" -type f -path "*/.githooks/*" | while IFS= read -r HOOK_FILE; do
         if grep -q "disabled> $HOOK_FILE" "${CURRENT_GIT_DIR}/.githooks.checksum"; then
             echo "Hook file is currently disabled at $HOOK_FILE"
             continue
@@ -873,10 +877,10 @@ remove_shared_hook_repo() {
         CURRENT_LIST=$(git config --global --get githooks.shared)
         NEW_LIST=""
 
-        IFS=",
-        "
+        IFS="$IFS_COMMA_NEWLINE"
 
         for SHARED_REPO_ITEM in $CURRENT_LIST; do
+            unset IFS
             if [ "$SHARED_REPO_ITEM" = "$SHARED_REPO_URL" ]; then
                 continue
             fi
@@ -886,6 +890,7 @@ remove_shared_hook_repo() {
             else
                 NEW_LIST="${NEW_LIST},${SHARED_REPO_ITEM}"
             fi
+            IFS="$IFS_COMMA_NEWLINE"
         done
 
         unset IFS
@@ -910,10 +915,10 @@ remove_shared_hook_repo() {
         CURRENT_LIST=$(grep -E "^[^#].+$" <"$(pwd)/.githooks/.shared")
         NEW_LIST=""
 
-        IFS=",
-        "
+        IFS="$IFS_COMMA_NEWLINE"
 
         for SHARED_REPO_ITEM in $CURRENT_LIST; do
+            unset IFS
             if [ "$SHARED_REPO_ITEM" = "$SHARED_REPO_URL" ]; then
                 continue
             fi
@@ -924,6 +929,7 @@ remove_shared_hook_repo() {
                 NEW_LIST="${NEW_LIST}
 ${SHARED_REPO_ITEM}"
             fi
+            IFS="$IFS_COMMA_NEWLINE"
         done
 
         unset IFS
@@ -1020,16 +1026,16 @@ list_shared_hook_repos() {
         esac
     done
 
-    IFS=",
-    "
-
     if [ -n "$LIST_GLOBAL" ]; then
         echo "Global shared hook repositories:"
 
         if [ -z "$(git config --global --get githooks.shared)" ]; then
             echo "  - None"
         else
+
+            IFS="$IFS_COMMA_NEWLINE"
             for LIST_ITEM in $(git config --global --get githooks.shared); do
+                unset IFS
 
                 set_shared_root "$LIST_ITEM"
                 if [ -d "$SHARED_ROOT/.git" ]; then
@@ -1048,7 +1054,10 @@ list_shared_hook_repos() {
                 else
                     echo "  - $NORMALIZED_NAME ($LIST_ITEM_STATE)"
                 fi
+
+                IFS="$IFS_COMMA_NEWLINE"
             done
+            unset IFS
         fi
     fi
 
@@ -1063,7 +1072,9 @@ list_shared_hook_repos() {
         else
             SHARED_REPOS_LIST=$(grep -E "^[^#].+$" <"$(pwd)/.githooks/.shared")
 
+            IFS="$IFS_COMMA_NEWLINE"
             echo "$SHARED_REPOS_LIST" | while read -r LIST_ITEM; do
+                unset IFS
 
                 set_shared_root "$LIST_ITEM"
 
@@ -1083,11 +1094,13 @@ list_shared_hook_repos() {
                 else
                     echo "  - $NORMALIZED_NAME ($LIST_ITEM_STATE)"
                 fi
+
+                IFS="$IFS_COMMA_NEWLINE"
             done
+            unset IFS
         fi
     fi
 
-    unset IFS
 }
 
 #####################################################
@@ -1146,10 +1159,11 @@ update_shared_hooks_in() {
     SHARED_REPOS_LIST="$1"
 
     # split on comma and newline
-    IFS=",
-    "
+    IFS="$IFS_COMMA_NEWLINE"
 
     for SHARED_REPO in $SHARED_REPOS_LIST; do
+        unset IFS
+
         mkdir -p ~/.githooks/shared
 
         set_shared_root "$SHARED_REPO"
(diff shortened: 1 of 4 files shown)


gabyx:
Jeah, of course.

gabyx:
Maybe lets add the tests later in another PR. We should actually run all tests against some whitespaces. I can probably test this by simply hacking the exec-tests.sh and looking for failed tests.

rycus86:
I'm more concerned about `IFS` on all platforms, so a duplicate test would have the benefit of testing it on all the platforms that are configured already.

gabyx:
Ah men: Shitty IFS crap is so dangerous:
```shell
   echo "$LOCAL_REPOSITORY_LIST" | while IFS= read -r EXISTING; do
        echo "Ex: $EXISTING"
        install_hooks_into_repo "$EXISTING_REPO_ROOT"
    done

```

**Output**
```shell
Do you want to install the hooks into existing repositories? [y/N] Where do you want to start the search? [/root]
LIST: /tmp/test045/001/.git
/tmp/test045/002/.git
with xargs:
f: /tmp/test045/001/.git
f: /tmp/test045/002/.git
Ex: /tmp/test045/001/.git  <<<<.---- WHY ONLY THE FIRST one??
Looks like you don't have a .githooks folder in the /tmp/test045/001 repository yet.
```
does not work, but if I comment out `install_hooks_into_repo` then the loop processes all `\n` delimited paths correctly, otherwise the 

I prefere something more robust instead: `xargs`

gabyx:
Also the decision to not have `'` quotes doesnt enlighten me much, since sometimes you want to control or restrict expansion, and this decision only makes it harder... at least for me...

gabyx:
Ah its because the install script mocks out `</dev/tty` which in my case reads my loop stdin... shit.

gabyx:
The cleanest way I found is the following in f9a214456
the while loop changes the stdinput which is bad and undefined behavior for successive calls.
I testet various methods (for loop, with IFS resetting, looks totally ugly) etc...

gabyx:
I added the test: `step-045w.sh` with whitespaces which would fail if this bugfix is not applied.

gabyx:
I replaced all tests to have whitespaces in a lot of paths (next two commits), this is definitely something we need to check and detect failure early, Thats why I suggest we use all these new tests (`test\d+` -> `test $1`)
They all pass and work 👍 . 
Thanks for reviewing this and merging this in.

install.sh
```diff
@@ -3545,10 +3552,16 @@ install_into_existing_repositories() {
...
     fi
 
-    for EXISTING in $LOCAL_REPOSITORY_LIST; do
+    # We dont want to change the std input through
+    # a while loop, so we redirect over descriptor 3
+    INPUT_PIPE=$(mktemp)
+    rm -rf "$INPUT_PIPE" >/dev/null 2>&1 || true
+    mkfifo "$INPUT_PIPE"
+    echo "$LOCAL_REPOSITORY_LIST" >"$INPUT_PIPE" &
+    while IFS= read -r EXISTING <&3; do
```
gabyx: Thats the best I could come up with which is POSIX...
If you have smoother way tell me, 
```shell
IFS="
"
for A in $A; do
   cmd-which-uses-std-in
done;
unset IFS
```
Does not work...

cli.sh
```diff
@@ -1147,7 +1147,8 @@ update_shared_hooks_in() {
 
     # split on comma and newline
     IFS=",
-    "
+"
```
gabyx: This is super crucial, as it splits also for whitespaces...

rycus86:
Sorry, I was unavailable for a few days, just looking at this now.
82 changed files though doesn't really help. :)
Can we first do the actual change and perhaps add one or two new tests rather than change all the existing ones?

On the actual whitespace splitting problem, I just tried this in `alpine`'s shell:
```
# find . -type f
./def ghi/test
./abc/test file
./jkl/sample
./jkl/mno/another file
# RLIST=$(find . -type f)
#
# this below is what you're trying to fix, right?
# for T in $RLIST; do echo "Item: $T"; done
Item: ./def
Item: ghi/test
Item: ./abc/test
Item: file
Item: ./jkl/sample
Item: ./jkl/mno/another
Item: file
# IFS=",
"
# for T in $RLIST; do echo "Item: $T"; done
Item: ./def ghi/test
Item: ./abc/test file
Item: ./jkl/sample
Item: ./jkl/mno/another file
```

Should we then just ensure all unquoted for loops are executed with the appropriate `IFS` set?

gabyx:
Correct. But setting the IFS is really a bit buggy. Because if we set it, then all commands inside the for loop have the same IFS, which is buggy (when we change things..)
I force pushed two commits, the Bugfix, and all tests with whitespaces...
Could you look at the changes first?
The tests are only whitespace replacements, nothing changed there.

gabyx:
The robust reference is: https://unix.stackexchange.com/questions/9496/looping-through-files-with-spaces-in-the-names

gabyx:
by the way:
image
 In this way I used:
```shell
 # We dont want to change the std input through
    # a while loop, so we redirect over descriptor 3
    INPUT_PIPE=$(mktemp)
    rm -rf "$INPUT_PIPE" >/dev/null 2>&1 || true
    mkfifo "$INPUT_PIPE"
    find "$START_DIR" -path "*templates/hooks/pre-commit.sample*" 2>/dev/null >"$INPUT_PIPE" &
    while IFS= read -r HIT <&3; do
```

gabyx:
I make these changes in the urge to use this tool on windows, but as you know windows is crap and also whitespaces in paths are used a lot there... Maybe we should also add an appveyor test or something like that :-), basically just installing gitforwindows and then run everything there :-)

rycus86:
Thanks! I had the ~~pain~~ privilege of working with Windows for a long time, I don't envy you... :)
For `SC2044`, my understanding is that `find | grep` is less buggy in a for loop than plain `find`s, not sure where I've got this from and could be wrong.
All the changes look OK to me, except the ones that now use temporary files and pipes and whatnot. Is there really nothing simpler we can do? Those are two loops only, even setting `IFS` multiple times would be simpler I think.

gabyx:
Jeah I dont like the pipe also not. Maybe we can just use the 
IFS=
and stick with if it works. Let me change it.

cli.sh
rycus86: Nice one, good catch!

install.sh
```diff
@@ -3388,7 +3387,11 @@ search_for_templates_dir() {
 search_pre_commit_sample_file() {
     START_DIR="$1"
 
-    for HIT in $(find "$START_DIR" 2>/dev/null | grep "templates/hooks/pre-commit.sample"); do
+    IFS="
+"
+    # shellcheck disable=SC2044
+    for HIT in $(find "$START_DIR" -path "*templates/hooks/pre-commit.sample*" 2>/dev/null); do
```
rycus86: Nit: this would match `pre-commit.sample.whatever` too. I don't think the trailing `*` hurts the `find` performance too much, but I'd drop it.

install.sh
```diff
@@ -3616,9 +3620,16 @@ install_into_existing_repositories() {
         LOCAL_REPOSITORY_LIST=$(echo "$LOCAL_REPOSITORY_LIST" | sort)
     fi
 
+    TEMP_IFS="
```
rycus86: Should this go like this?
```
LOOP_IFS="
"
PREVIOUS_IFS="$IFS"
IFS="$LOOP_IFS"
for EXISTING in $LOCAL_REPOSITORY_LIST; do
    unset IFS
    ...
    IFS="$LOOP_IFS"
done
IFS="$PREVIOUS_IFS"
```

tests/exec-steps.sh
```diff
@@ -20,6 +20,8 @@ for STEP in /var/lib/tests/step-*.sh; do
         mkdir -p /usr/share/git-core/templates/hooks
         rm -rf /usr/share/git-core/templates/hooks/*
     fi
+
+    rm -rf ~/test*
```
rycus86: Is this related to the whitespace in all the test names? If so, please revert for now.

rycus86:
OK, I think this almost looks ready go. :+1: 
Added some comments for some style changes, plus please revert all the test changes for now.
One (new) whitespace test would be nice, but I'm happy to add one myself after merging this PR.
And I'll also add a new Travis job to automatically change all the paths in the test scripts to have a space, and test the whitespaces in general that way.

rycus86:
Oh `step-006` was failing on Travis with your latest changes. :thinking:

gabyx:
> OK, I think this almost looks ready go. 👍
> Added some comments for some style changes, plus please revert all the test changes for now.
> One (new) whitespace test would be nice, but I'm happy to add one myself after merging this PR.
> And I'll also add a new Travis job to automatically change all the paths in the test scripts to have a space, and test the whitespaces in general that way.

That sounds perfect.
I do the review in moment.
What do you mean by step-006 was failing? I check the tests when I ve done the review. Lets see. Thanks :-)

install.sh
gabyx: I think we should always unset the IFS. not setting it to the previous one.
Meaning when we want to use special behavior, we set the IFS only in the scope we need it, and afterward reset it to `unset IFS` meaning standard wordsplitting is used "<tab><space><newline>"

rycus86: Ok, sounds good.

tests/exec-steps.sh
gabyx: No it isn't, these are some artefacts from tests which write to the home directory, as I found :-) 
I 'd leave it :-)

gabyx:
Can we enumerate
```LOOP_IFS_<linenumber>``` and have a pre-commit script which reassigns the line numbers always...?

**Comment by rycus86** [comment]
> Can we enumerate
> `LOOP_IFS_<linenumber>` and have a pre-commit script which reassigns the line numbers always...?

Why is this a concern?

**Comment by gabyx** [comment]
Because of nestet for loops: Using a `LOOP_IFS=...` might  might be reset if another shell function iside the loop resets LOOP_IFS to something else...
Maybe define globals some IFS.
IFS_COMMA_NEWLINE
IFS_NEWLINE
and only use them to reassign IFS
```
IFS="$IFS_COMMA_NEWLINE"
for ...
unset IFS
...
IFS="$IFS_COMMA_NEWLINE"
done

rycus86:
> Because of nestet for loops: Using a `LOOP_IFS=...` might might be reset if another shell function iside the loop resets LOOP_IFS to something else...

OK, what I meant to ask is: is this a real concern here? Does it happen already? How hard/easy it is to make it happen?

If it's not super easy to mess up, I'd not worry about it for now.

gabyx:
See my latest commit. global IFS to use in LOOPS, that should fix also potential future problems.
It think this way its kind of more robust.

cli.sh
```diff
@@ -1064,6 +1072,7 @@ list_shared_hook_repos() {
             SHARED_REPOS_LIST=$(grep -E "^[^#].+$" <"$(pwd)/.githooks/.shared")
 
             echo "$SHARED_REPOS_LIST" | while read -r LIST_ITEM; do
```
rycus86: Missing `IFS="$IFS_COMMA_NEWLINE"` before the loop?

tests/exec-steps.sh
rycus86: OK

rycus86:
Nice, looks good to me! :+1: 
Let's wait for Travis then I'll merge it in.

rycus86:
Hm, somethings seems to have broken around shared repo lists, see steps `082` and `092`.

gabyx:
Seen it :-) Will investigate.

cli.sh
```diff
@@ -66,6 +66,10 @@ set_main_variables() {
     if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
         CURRENT_GIT_DIR=".git"
     fi
+
+    # Global IFS for loops
+    IFS_COMMA_NEWLINE="
```
rycus86: Oh, I missed this. You say `COMMA_NEWLINE` but doesn't actually include `,`

gabyx: damm, thx!

gabyx: At least we know that some test fail like that :)

rycus86:
> Hm, somethings seems to have broken around shared repo lists, see steps `082` and `092`.

OK, found the problem I think, left a comment.

gabyx:
Thx!

rycus86:
> Thx!

Thank you! :)

Issue #17, opened 2019-07-03
What about git LFS?


AlexGoris-Cashforce:
I'm preparing a pull request with some changes which would make this project better suited for our internal use, however I've stumbled across an issue:

we have some repositories using git-lfs (large file storage), which als uses hooks to push lfs objects (matched through filters specific in `.gitattributes`) to the LFS storage which corresponds to the current repository.
These hooks effectively get overwritten when installing this project. So should we include the LFS hooks in `.githooks/*` on all repo's which use LFS, or is there a better/other solution?

rycus86:
Hm, I haven't really used git-lfs yet, so I'm not sure how this should work. We could add a test for it and see what happens, I'd imagine it might need some fixes.

Thanks for bringing this here!

gabyx:
We also use git lfs? How does `git lfs install` install files in the `.git/hooks` directory?
I have not seen any?. Isnt git-lfs invoked by git itself through the settings in the global config (installed with `git lfs install`) For every commit git pushed the files itself to LFS , the ones which match the `.gitattributes` files...??

rycus86:
If either of you could set up a sample repo to test on, I could look into adding a test for it.

AlexGoris-Cashforce:
I've set up a test repo with a 10MB.zip file added through git-lfs
https://github.com/AlexGoris-Cashforce/githooks-lfs-test

The documentation around git-lfs seems to be a bit iffy...
In the [manpage ](https://github.com/git-lfs/git-lfs/blob/master/docs/man/git-lfs-install.1.ronn)you can read that `git lfs install`, amon other things:
>Install a pre-push hook to run git-lfs-pre-push(1) for the current repository, if run from inside one. If "core.hooksPath" is configured in any Git configuration (and supported, i.e., the installed Git version is at least 2.9.0), then the pre-push hook will be installed to that directory instead.

Now it does more than just installing a pre-push hooks, if I run `git lfs install` in a clean repo, I get 4 new hooks:
* post-checkout
* post-commit
* post-merge
* pre-push

[...]

PR #30, opened 2019-07-30
Support for running Git LFS hooks :green_heart:

rycus86:
This is my attempt to solve #17 
@gabyx @AlexGoris-Cashforce can you please have a look?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   - TEST_INSTANCE=test-alpine-curl
   - TEST_INSTANCE=test-alpine-wget
   - TEST_INSTANCE=test-alpine-user
+  - TEST_INSTANCE=test-alpine-lfs
   - TEST_INSTANCE=test-debian
   - TEST_INSTANCE=test-debian-user
   - TEST_INSTANCE=test-ubuntu
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1907.311448-fee6c0
+# Version: 1908.012129-2e301c
 
 #####################################################
 # Execute the current hook,
@@ -19,6 +19,7 @@ process_git_hook() {
     export_staged_files
     check_for_updates_if_needed
     execute_old_hook_if_available "$@" || return 1
+    execute_lfs_hook_if_appropriate "$@" || return 1
     execute_global_shared_hooks "$@" || return 1
     execute_local_shared_hooks "$@" || return 1
     execute_all_hooks_in "$(pwd)/.githooks" "$@" || return 1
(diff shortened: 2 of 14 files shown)


rycus86:
Unfortunately, Git LFS will still print a warning on install that the existing hook is not theirs, but otherwise should be working as far as I can tell.

gabyx:
Nice. 
Maybe we should add a test. which without installing githooks inits a repo, and compares the output to the hook switch we are using inside base-template.

gabyx:
> Unfortunately, Git LFS will still print a warning on install that the existing hook is not theirs, but otherwise should be working as far as I can tell.

What is the output of that?

rycus86:
> What is the output of that?

It says:
```
$ git lfs install
Hook already exists: pre-push

        #!/bin/sh
        # Base Git hook template from https://github.com/rycus86/githooks
        #
        # It allows you to have a .githooks folder per-project that contains
        # its hooks to execute on various Git triggers.
        #
        # Version: 1907.110919-c0438b

        # Execute the current hook,
        #   that in turn executes the hooks in the repo.
        #
        # Returns:
        process_git_hook() {
        are_githooks_disabled && return 0
        set_main_variables
        export_staged_files
        check_for_updates_if_needed
        execute_old_hook_if_available "$@" || return 1
        execute_global_shared_hooks "$@" || return 1
        execute_local_shared_hooks "$@" || return 1
        execute_all_hooks_in "$(pwd)/.githooks" "$@" || return 1
        }

        # Checks if Githooks is completely disabled
        #   for the current repository or globally.
        #   This can be done with Git config or using
        #   the ${GITH

To resolve this, either:
  1: run `git lfs update --manual` for instructions on how to merge hooks.
  2: run `git lfs update --force` to overwrite your hook.
```

rycus86:
> Maybe we should add a test. which without installing githooks inits a repo, and compares the output to the hook switch we are using inside base-template.

You mean 1) install LFS first with its hooks, then 2) install Githooks and 3) check Githooks hooks are run with also triggers `git-lfs` ?

gabyx:
Ah that one, ah easy. 
```git lfs install``` is done on a machine before githooks is installed. So it should be fine...

For the tests I mean:
- `git init` should install all lfs hooks (when you have done `git lfs install`
- Check the automatically installed  lfs original hooks in `.git/hooks/*`  against our function `execute_lfs_hook_if_appropriate`.
Making sure we have lfs in the same hooks as Lfs automatically installs.

**Comment by rycus86** [comment]
Oh OK, I think I see what you're saying. I'll try to have a look tomorrow/this week, time permitting. Thanks for the comments!

**Comment by gabyx** [comment]
Second test.
- `git install lfs`
- `./install.sh`
- `git init`
- `git commit`
- `check if post-commit lfs has been run` maybe moch `git-lfs` by  config: 
```ini
[alias] 
lfs = our-mock-script
```

gabyx:
Try to add the first test case.

rycus86:
> Try to add the first test case.

I'm also looking into this, just got it working. Will push in a minute.

gabyx:
See two commits on the branch : https://github.com/gabyx/githooks/tree/lfs-support

rycus86:
> See two commits on the branch : https://github.com/gabyx/githooks/tree/lfs-support

Yep, similar approach, different implementation. :)

rycus86:
OK, added both tests now. Can you please have a look @gabyx ?
Do you have anything extra on your branch you'd want me to consider?

gabyx:
Plz have a look at this, can we include that.
My branch commit : e6f01e9

rycus86:
> Plz have a look at this, can we include that.
> My branch commit : [e6f01e9](https://github.com/rycus86/githooks/commit/e6f01e9b7c9ef429de4063b61afce1acdee66ad1)

Why not just have a `.githooks/pre-commit/require-lfs` and provide your own script?
I'm not against the extension, just wondering what's the added value? Less repetition? Easier setup?

gabyx:
Ah jeah of course :). We should write a Readme, to include whats happning with LFS for the user!

gabyx:
The PR is nice, so with that and #29 I can finally test the setup for the company wide Git Installation. Thx!

rycus86:
> Ah jeah of course :). We should write a Readme, to include whats happning with LFS for the user!

Added #31 to track it.

rycus86:
> Plz have a look at this, can we include that.
> My branch commit : [e6f01e9](https://github.com/rycus86/githooks/commit/e6f01e9b7c9ef429de4063b61afce1acdee66ad1)

Can you check the PR now? I added support for the `.githooks/.lfs-required` file.

rycus86:
Sooo, bit of a complication with `.lfs-required`. Most LFS hooks are `post-*` which can't fail the operation since they run after that. So that leaves us with `pre-push`. Is that good enough to fail there?
Would you think maybe we should check elsewhere too? Like `pre-commit` ?

gabyx:
`git lfs` hooks are:
```shell
Add the following to .git/hooks/pre-push:

        #!/bin/sh
        command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, 
remove this hook by deleting .git/hooks/pre-push.\n"; exit 2; }
        git lfs pre-push "$@"

Add the following to .git/hooks/post-checkout:

        #!/bin/sh
        command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, 
remove this hook by deleting .git/hooks/post-checkout.\n"; exit 2; }
        git lfs post-checkout "$@"

Add the following to .git/hooks/post-commit:

        #!/bin/sh
        command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, 
remove this hook by deleting .git/hooks/post-commit.\n"; exit 2; }
        git lfs post-commit "$@"

Add the following to .git/hooks/post-merge:

        #!/bin/sh
        command -v git-lfs >/dev/null 2>&1 || { echo >&2 "\nThis repository is configured for Git LFS but 'git-lfs' was not found on your path. If you no longer wish to use Git LFS, 
remove this hook by deleting .git/hooks/post-merge.\n"; exit 2; }
        git lfs post-merge "$@"
```
Strange that they do not provide `git lfs pre push "$@ || exit 1`
Hm... do you know why?

gabyx:
Maybe only fail at pre-push, yes. And print a warning for the others if something happened? Why not also `exit 1?`

gabyx:
For me that PR looks good, merge when ever you are ready. Thx!

rycus86:
I think it does exit with non-zero and we will too. But as https://git-scm.com/docs/githooks#_post_commit states:
> This hook is meant primarily for notification, and __cannot affect__ the outcome of git commit.

rycus86:
I'll try to wrap this up this evening and merge it in.

rycus86:
OK, added one more test, and will merge this once Travis is happy. Some good news: `post-checkout` fails both the initial `git clone` and also further `git checkout` if the `.githooks/.lfs-required` file is there but `git-lfs` is not found. :tada:

gabyx:
ah perfect!

Von meinem iPhone gesendet

gabyx:
But strange when you clone you clone the repo file `.githooks/.lfs-required` file and afterwards its fails? 

Von meinem iPhone gesendet

gabyx:
got it :)

Von meinem iPhone gesendet

PR #32, opened 2019-08-02
Readme :zap:

gabyx:


gabyx:
Also some small trivial code fixes

README.md
```diff
@@ -43,12 +44,13 @@ If a file is executable, it is directly invoked, otherwise it is interpreted wit
 Hooks related to `commit` events will also have a `${STAGED_FILES}` environment variable set, that is the list of staged and changed files (according to `git diff --cached --diff-filter=ACMR --name-only`), one per line and where it makes sense (not `post-commit`). If you want to iterate over them, and expect spaces in paths, you might want to set `IFS` like this.
 
 ```shell
-IFS="
-"
-
+IFS="$IFS_NEWLINE"
```
rycus86: We don't say what is `IFS_NEWLINE` here.
But also, I'm not too sure if this extra complexity is needed in this code example?
Unless something resets `IFS` inside the loop, you don't need this.

README.md
```diff
@@ -77,6 +79,22 @@ The supported hooks are listed below. Refer to the [Git documentation](https://g
 - `post-rewrite`
 - `sendemail-validate`
 
+## GIT Large File System Support
```
rycus86: How about:

```
## Git Large File Storage support

If the user has installed [Git Large File Storage](https://git-lfs.github.com/) (`git-lfs`) by calling 
`git lfs install` globally or locally for a repository only, `git-lfs` installs 4 hooks when initializing (`git init`) or cloning (`git clone`) a repository:

- `post-checkout`
- `post-commit`
- `post-merge`
- `pre-push`

Since `githooks` overwrites the hooks in `.git/hooks`, it will also run all *Git LFS* hooks internally if the `git-lfs` executable is found on the system path. You can enforce having `git-lfs` installed on the system by placing a `./githooks/.lfs-required` file inside the repository, then if `git-lfs` is missing, a warning is shown and the hook will exit with code `1`. For some `post-*` hooks this does not mean that the outcome of the git command can be influenced even tough the exit code is `1`, for example `post-commit` hooks can't fail commits. A clone of a repository containing this file might still work but would issue a warning and exit with code `1`, a push - however - will fail if `git-lfs` is missing.

It is advisable for repositories using *Git LFS* to also have a pre-commit hook (e.g. `examples/lfs/pre-commit`) checked in which enforces a correct installation of *Git LFS*.
```

README.md
```diff
@@ -97,6 +115,13 @@ $ git hooks shared list --with-url
 ```
 
 The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `~/.githooks.shared` folder, and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
+The additional global configuration parameter `githooks.failOnNotExistingSharedHooks`:
```
rycus86: How about this?

```
An additional global configuration parameter `githooks.failOnNotExistingSharedHooks` makes hooks fail with an error if any shared hook configured in `.shared` is missing, meaning `git hooks update` has not yet been called.
Note that shared hooks are automatically updated on clone.
```

Also, this reminds me, that we should have added a `git hooks config [enable|disable] fail-on-non-existing-shared-hooks` to the cli.

examples/lfs/pre-commit
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
rycus86: This won't actually fail the commit. Did you want to add `exit 1` in appropriate places?

README.md
gabyx: ah jeah, sure, this is code in the hook which we execute in a new shell...

examples/lfs/pre-commit
```diff
@@ -0,0 +1,13 @@
...
+# LFS Pre-commit hook to test for a properly working Git LFS.
+
+command -v git-lfs >/dev/null 2>&1 || echo "This repository needs Git LFS but \`git-lfs\` was not found on your path."
+
+if [ "$(git config --get filter.lfs.required)" != "true" ] ||
```
rycus86: Maybe like this?

```
lfs_config_check_failed() {
    echo "! Git LFS filter are not installed properly!"
    echo "  Run: \`git lfs install\` to correct this."
    exit 1
}

[ "$(git config --get filter.lfs.required)" = "true" ] || lfs_config_check_failed
git config --get filter.lfs.clean | grep -q "git-lfs" || lfs_config_check_failed
git config --get filter.lfs.smudge | grep -q "git-lfs" || lfs_config_check_failed
git config --get filter.lfs.process | grep -q "git-lfs" || lfs_config_check_failed
```

README.md
gabyx: Can you just pull and merge your stuff in :-), your native speaker which is in all ways better :-)

README.md
gabyx: Hm... yes. I would use `failOnNotExistingSharedHooks` same as in the config.

examples/lfs/pre-commit
gabyx: I do that.

README.md
rycus86: Sure, I just want to keep your name on the `git blame` for credits. :)
FYI, I am not a native speaker. :)

README.md
gabyx: No its good as you propsed.

README.md
gabyx: I do it.

rycus86:
Thanks a lot for this! I fixed up some bits and merged. :tada:

PR #34, opened 2019-08-03
Errors to stderr

gabyx:


base-template.sh
```diff
@@ -138,8 +138,8 @@ execute_lfs_hook_if_appropriate() {
     if [ "$GIT_LFS_AVAILABLE" = "true" ]; then
         git lfs "$HOOK_NAME" "$@" || return 1
     elif [ "$REQUIRES_LFS_SUPPORT" = "true" ]; then
-        echo "This repository requires Git LFS, but \`git-lfs\` was not found on your PATH."
-        echo "If you no longer want to use Git LFS, remove the \`.githooks/.lfs-required\` file."
+        echo "! This repository requires Git LFS, but \`git-lfs\` was not found on your PATH." >&2
+        echo "!  If you no longer want to use Git LFS, remove the \`.githooks/.lfs-required\` file." >&2
```
rycus86: As a general theme:
- please keep the `!` on the start of the first line only (no need to have all lines start with `!`)
- please don't change the indentation on lines, they should start where the previous did in most cases
Example:

```
! This is the first line of error
  This is some text explaining what to do, etc.
```

base-template.sh
```diff
@@ -452,7 +452,7 @@ update_shared_hooks_if_appropriate() {
                 PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
-                    echo "! Update failed, git pull output:"
+                    echo "! Update failed, git pull output:" >&2
                     echo "$PULL_OUTPUT"
```
rycus86: Redirect this to `>&2` as well.

base-template.sh
```diff
@@ -461,7 +461,7 @@ update_shared_hooks_if_appropriate() {
                 CLONE_OUTPUT=$(git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
-                    echo "! Clone failed, git clone output:"
+                    echo "! Clone failed, git clone output:" >&2
                     echo "$CLONE_OUTPUT"
```
rycus86: Redirect this to `>&2` as well.

base-template.sh
```diff
@@ -295,7 +295,7 @@ is_trusted_repo() {
 
         # shellcheck disable=SC2181
         if [ $TRUST_ALL_RESULT -ne 0 ]; then
-            echo "! This repository wants you to trust all current and future hooks without prompting"
+            echo "! This repository wants you to trust all current and future hooks without prompting" >&2
```
rycus86: This is a prompt rather than a failure, please keep it on stdout.

base-template.sh
gabyx: Ah I wasnt sure, so now its clear, I though you wanted to distinguise further error message from each other by indenting...

gabyx: So one "!" marks one error! I correct that overall

base-template.sh
gabyx: The prompt comes afterwards I though (printf)

rycus86:
Thanks a lot!

Issue #40, opened 2019-08-05
Install folder customization


gabyx:
It would be really nice to take up again the install folder again, since it would make usage of the following scenario much easier, as well as it is more powerful:

When I install githooks on Windows it will be installed to `C:/Users/.../.githooks`.
The git config ins Ubuntu Subsystem `/root/.gitconfig` would then look something like:
```
[githooks]        
        failOnNonExistingSharedHooks = true
        python = /.../.../AppData/Roaming/.../python/python.exe # my custom stuff...
[alias]
        hooks = !/mnt/c/Users/gabriel.nuetzi/.githooks/bin/githooks
```
When you wanna do some git manipulation in *windows ubuntu subystem*.
You end up reinstalling shared hooks somewhere else `/root/.githooks/...`.
This is quite cumbersome. If githooks would know the installation directory (extract from alias), using githooks from bash in the subsystem behaves identically as on windows git-bash.exe.

It would be nice, to have the already proposed `set_install_dir` in place and replace all occurances `~/.githooks` with a proper install dir. and also provide a `--prefix` option in `install.sh`.

rycus86:
So the problem here is that the user in Git Bash is different than in WSL? That sounds like a general configuration issue, not specific to Githooks.
I'm not against adding an extra flag to the install script, it's just cumbersome and one more flag to confuse users.

rycus86:
> If githooks would know the installation directory (extract from alias), using githooks from bash in the subsystem behaves identically as on windows git-bash.exe.

That sounds like relying on a side effect. If we do this, we should set a global configuration value that points to the install dir (which I thought we have already), and use that. Is the problem here that the global Git config file will be a different one?
If yes, do you have to set up all your Git config twice? Like committer name, email, etc.

gabyx:
Yes the config files need to be setup also in WSL under `root/.gitconfig`
If I wanna use the same `alias.hooks` path, it wont work properly since the install directory is different in the subsystem and windows install...

> If we do this, we should set a global configuration value that points to the install dir (which I thought we have already)

Where is this variable? We dont have this, I made once some changes in this direction, but during review we dropped this...

gabyx:
I would make a PR with a setting
```ini
[githooks]
 installDir = "...."
```
and `--prefix` for the installation. 
Would that be of value?

rycus86:
We can talk about it.
Is the problem here not that the `~/.gitconfig` file will be different?
Because in that case we can't really set a config value for both places.

[...]

PR #42, opened 2019-08-14
Customized install prefix

gabyx:
#40

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ $ git hooks shared list --with-url
 ...
 ```
 
-The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `~/.githooks.shared` folder, and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
+The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default)), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
 An additional global configuration parameter `githooks.failOnNonExistingSharedHooks` makes hooks fail with an error if any shared hook configured in `.shared` is missing, meaning `git hooks update` has not yet been called. See `git hooks config [enable|disable] fail-on-non-existing-shared-hooks` in the [command line helper](https://github.com/rycus86/githooks/blob/master/docs/command-line-tool.md) tool documentation for more information.
 Note that shared hooks are automatically updated on clone.
 
@@ -178,6 +178,8 @@ $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --dry-run
 ```
 
 You can also run the installation in non-interactive mode with the command below. This will determine an appropriate template directory (detect and use the existing one, or use the one passed by `--template-dir`, or use a default one), install the hooks automatically into this directory, and enable periodic update checks.
+
+The global installation prefix defaults to `$HOME` but can be changed by passing `--prefix <install-prefix>` to the `install.sh` script.
 
 ```shell
 $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --non-interactive
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1908.082043-b503e9
+# Version: 1908.220946-65f51e
 
 #####################################################
 # Execute the current hook,
@@ -46,6 +46,28 @@ are_githooks_disabled() {
 }
 
 #####################################################
+# Sets the ${INSTALL_DIR} variable
+
+# Returns: 0 if success, 1 otherwise
+#####################################################
+load_install_dir() {
+
+    INSTALL_DIR=$(git config --global --get githooks.installDir)
+
+    #shellcheck disable=SC2181
+    if [ -z "${INSTALL_DIR}" ]; then
+        # install dir not defined, use default
+        INSTALL_DIR=~/".githooks"
+    elif [ ! -d "$INSTALL_DIR" ]; then
+        echo "! Githooks installation is corrupt! " >&2
+        echo "  Install directory \`${INSTALL_DIR}\` is missing." >&2
+        INSTALL_DIR=~/".githooks"
+        echo "  Fallback to default directory \`${INSTALL_DIR}\`" >&2
+        echo "  Please run the Githooks install script again to fix it." >&2
+    fi
+}
+
+#####################################################
 # Set up the main variables that
 #   we will throughout the hook.
 #
@@ -54,8 +76,7 @@ are_githooks_disabled() {
 # Resets the ${ACCEPT_CHANGES} variable
 # Sets the ${CURRENT_GIT_DIR} variable
 #
-# Returns:
-#   None
+# Returns: None
 #####################################################
 set_main_variables() {
     HOOK_NAME=$(basename "$0")
@@ -66,6 +87,8 @@ set_main_variables() {
     if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
         CURRENT_GIT_DIR=".git" # reset to a sensible default
     fi
+
+    load_install_dir
 
     # Global IFS for loops
     IFS_COMMA_NEWLINE=",
@@ -419,7 +442,7 @@ set_shared_root() {
     NORMALIZED_NAME=$(echo "$1" |
         sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
         sed -E "s/[^a-zA-Z0-9]/_/g")
-    SHARED_ROOT=~/.githooks/shared/"$NORMALIZED_NAME"
+    SHARED_ROOT="$INSTALL_DIR/shared/$NORMALIZED_NAME"
 }
 
 #####################################################
@@ -444,7 +467,7 @@ update_shared_hooks_if_appropriate() {
 
         for SHARED_REPO in $SHARED_REPOS_LIST; do
             unset IFS
-            mkdir -p ~/.githooks/shared
+            mkdir -p "$INSTALL_DIR/shared"
 
             set_shared_root "$SHARED_REPO"
 
@@ -475,7 +498,7 @@ update_shared_hooks_if_appropriate() {
 
 #####################################################
 # Execute the shared hooks in the
-#   ~/.githooks/shared directory.
+#   $INSTALL_DIR/shared directory.
 #
 # Returns:
 #   1 in case a hook fails, 0 otherwise
@@ -607,12 +630,12 @@ should_run_update_checks() {
 #   `$1`
 #
 # Returns:
-#   0 and "~/.githooks/tools/$1/run"
+#   0 and "$INSTALL_DIR/tools/$1/run"
 #   1 and "" otherwise
 #####################################################
 get_tool_script() {
-    if [ -f ~/".githooks/tools/$1/run" ]; then
-        echo ~/".githooks/tools/$1/run" && return 0
+    if [ -f "$INSTALL_DIR/tools/$1/run" ]; then
+        echo "$INSTALL_DIR/tools/$1/run" && return 0
     fi
     return 1
 }
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1908.082043-b503e9
+# Version: 1908.220946-65f51e
 
 #####################################################
 # Prints the command line help for usage and
@@ -53,19 +53,44 @@ print_help_header() {
 }
 
 #####################################################
+# Sets the ${INSTALL_DIR} variable
+
+# Returns: 0 if success, 1 otherwise
+#####################################################
+load_install_dir() {
+
+    INSTALL_DIR=$(git config --global --get githooks.installDir)
+
+    #shellcheck disable=SC2181
+    if [ $? -ne 0 ]; then
+        # install dir not defined, use default
+        INSTALL_DIR=~/".githooks"
+    elif [ ! -d "$INSTALL_DIR" ]; then
+        echo "! Githooks installation is corrupt! " >&2
+        echo "  Install directory \`${INSTALL_DIR}\` is missing." >&2
+        INSTALL_DIR=~/".githooks"
+        echo "  Fallback to default directory \`${INSTALL_DIR}\`" >&2
+        echo "  Please run the Githooks install script again to fix it." >&2
+    fi
+
+    return 0
+}
+
+#####################################################
 # Set up the main variables that
 #   we will throughout the hook.
 #
 # Sets the ${CURRENT_GIT_DIR} variable
 #
-# Returns:
-#   None
+# Returns: None
 #####################################################
 set_main_variables() {
     CURRENT_GIT_DIR=$(git rev-parse --git-common-dir 2>/dev/null)
     if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
         CURRENT_GIT_DIR=".git"
     fi
+
+    load_install_dir
 
     # Global IFS for loops
     IFS_COMMA_NEWLINE=",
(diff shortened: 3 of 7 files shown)


base-template.sh
```diff
@@ -67,6 +67,8 @@ set_main_variables() {
         CURRENT_GIT_DIR=".git" # reset to a sensible default
     fi
 
+    INSTALL_DIR=$(git config --global githooks.installDir)
```
rycus86: I'd check with every usage if `[ -z "$INSTALL_DIR" ]` and set it to `$HOME` if it is. Might be worth extracting it into a function.

I wonder why tests that don't install first didn't fail. When this config is missing, the path would become relative to root. Sounds like I have some missing coverage here. :(

rycus86: Oh if this is the common variable setup function then we don't need to extract setting this up further, just an extra if here. Please do the same in the other files too.

install.sh
```diff
@@ -3256,6 +3262,13 @@ parse_command_line_arguments() {
             SINGLE_REPO_INSTALL="yes"
         elif [ "$p" = "--skip-install-into-existing" ]; then
             SKIP_INSTALL_INTO_EXISTING="yes"
+        elif [ "$prev_p" = "--prefix" ] && (echo "$p" | grep -qvE '^\-\-.*'); then
+            # Allow user to pass prefered install prefix
+            INSTALL_DIR="$p"
+            if [ ! -d "$INSTALL_DIR" ]; then
+                echo "! Install prefix '$INSTALL_DIR' does not exist!"
```
rycus86: No quotes please, we don't quote paths elsewhere either - I thin.

base-template.sh
gabyx: I will check during `set_main_variables`

gabyx: > Sounds like I have some missing coverage here. :(
Hm... strange

install.sh
```diff
@@ -3620,16 +3633,17 @@ setup_hook_templates() {
...
 install_command_line_tool() {
-    mkdir -p "$HOME/.githooks/bin" &&
-        echo "$CLI_TOOL_CONTENT" >"$HOME/.githooks/bin/githooks" &&
-        chmod +x "$HOME/.githooks/bin/githooks" &&
-        git config --global alias.hooks "!$HOME/.githooks/bin/githooks" &&
+    mkdir -p "$INSTALL_DIR/.githooks/bin" &&
+        echo "$CLI_TOOL_CONTENT" >"$INSTALL_DIR/.githooks/bin/githooks" &&
+        chmod +x "$INSTALL_DIR/.githooks/bin/githooks" &&
+        git config --global alias.hooks "!$INSTALL_DIR/.githooks/bin/githooks" &&
+        git config --global githooks.installDir "$INSTALL_DIR" &&
```
rycus86: Can we set this somewhere else?
It doesn't feel like it belongs in the `install_command_line_tool` function.
OK if you can't find a better place, just doesn't feel like this is the logical place for it.

gabyx:
Can I also change : `githooks.previous.searchdir` to `githooks.previous.searchDir`. Potentially removing the old variable in `uninstall.sh`?

rycus86:
> Can I also change : `githooks.previous.searchdir` to `githooks.previous.searchDir`. Potentially removing the old variable in `uninstall.sh`?

I'd leave it as it is, so it keeps working during updates.

gabyx:
This also adresses how you provide updates when we provide a new version. We dont have a version number. Every mistake :-) say typo, we make in the config is hard to correct =). 

Tests are noe failing where they should. The test-rule does not work yet, since the `cli.sh help` does crash since the install dir is not set. I have to make this more agnostic... Maybe make the help
Fail as early as possible is a good thing... Testing everywhere if the install dir is feasible is cumbersome and only puts weight on the performance...

gabyx:
Ok, I made the `cli.sh` a bit more agnostic.  First handling the `help` case, and afterwards the install directory is checked if its missing -> abort.

gabyx:
We still need to tweak the tests which do not install githooks.

gabyx:
Ok, now I think its ready for another review.

tests/exec-steps.sh
```diff
@@ -25,6 +25,10 @@ for STEP in /var/lib/tests/step-*.sh; do
     rm -rf ~/.githooks
     rm -rf /tmp/*
 
+    # Fake a default installation directory (for tests which do not install)
+    mkdir -p ~/.githooks
+    git config --global githooks.installDir ~/.githooks
```
rycus86: Please don't set this here, as this would make all tests **not** test the case when we don't have this set up.

install.sh
```diff
@@ -3920,6 +3997,21 @@ setup_shared_hook_repositories() {
...
 }
 
+############################################################
+# Finalizes the installation.
+#   Sets the global installation directory.
+#
+# Returns:
+#   None
+############################################################
+finalize_installation() {
```
rycus86: I'd set this up early, after we parsed the arguments, we know if we'll use the default or the prefix, so can be set right after that point.

install.sh
```diff
@@ -3256,12 +3319,26 @@ parse_command_line_arguments() {
...
+        elif [ "$prev_p" = "--prefix" ] && (echo "$p" | grep -qvE '^\-\-.*'); then
+            # Allow user to pass prefered install prefix
+            INSTALL_DIR="$p"
+            # Try to see if the path is given with a tilde
+            TILDE_REPLACED=$(echo "$INSTALL_DIR" | awk 'gsub("~", "'"$HOME"'", $0)')
+            if [ -n "$TILDE_REPLACED" ]; then
+                INSTALL_DIR="$TILDE_REPLACED"
+            fi
+            if [ ! -d "$INSTALL_DIR" ]; then
+                echo "! Install prefix '$INSTALL_DIR' does not exist!"
```
rycus86: `>&2` and drop the quotes please.

install.sh
```diff
@@ -3232,6 +3275,25 @@ execute_installation() {
...
     fi
+
+    finalize_installation || return 1
+}
+############################################################
+# Sets the install directory.
+#
+# Returns: None
+############################################################
+set_install_dir() {
```
rycus86: maybe `load_install_dir()` would be better, it would signal that this is setting the variable based on an existing setting.

install.sh
```diff
@@ -3232,6 +3275,25 @@ execute_installation() {
...
+    # First check if we have already a install directory set
+    INSTALL_DIR=$(git config --global githooks.installDir)
```
rycus86: I would:
- first check if `$INSTALL_DIR` is empty and set it to `$HOME/.githooks` if it is and `return`
- check if the directory exists and return
- fall back to `$HOME/.githooks` otherwise

install.sh
```diff
@@ -3232,6 +3275,25 @@ execute_installation() {
...
+    if [ ! -d "$INSTALL_DIR" ]; then
+
+        [ -z "$INSTALL_DIR" ] &&
+            echo "! Configured install directory \`$INSTALL_DIR\` not existing." >&2 &&
```
rycus86: We don't need quotes/backticks around paths here either.

install.sh
```diff
@@ -3256,12 +3319,26 @@ parse_command_line_arguments() {
...
             SKIP_INSTALL_INTO_EXISTING="yes"
+        elif [ "$prev_p" = "--prefix" ] && (echo "$p" | grep -qvE '^\-\-.*'); then
+            # Allow user to pass prefered install prefix
+            INSTALL_DIR="$p"
+            # Try to see if the path is given with a tilde
+            TILDE_REPLACED=$(echo "$INSTALL_DIR" | awk 'gsub("~", "'"$HOME"'", $0)')
+            if [ -n "$TILDE_REPLACED" ]; then
+                INSTALL_DIR="$TILDE_REPLACED"
+            fi
+            if [ ! -d "$INSTALL_DIR" ]; then
```
rycus86: Do we actually need to fail here?
Most (or hopefully all) code that will use this does an `mkdir -p`, right?

README.md
```diff
@@ -110,7 +110,7 @@ $ git hooks shared list --with-url
 ...
 ```
 
-The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `~/.githooks.shared` folder, and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
+The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `<install-dir>/.githooks.shared` folder, and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
```
rycus86: Maybe add the default here, so like:
> These repositories will be checked out into the `<install-dir>/.githooks.shared` folder (`~/.githooks.shared` by default), and are updated...

README.md
```diff
@@ -179,6 +179,8 @@ $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --dry-run
 
 You can also run the installation in non-interactive mode with the command below. This will determine an appropriate template directory (detect and use the existing one, or use the one passed by `--template-dir`, or use a default one), install the hooks automatically into this directory, and enable periodic update checks.
 
+The global installation folder defaults to `~/.githooks` but can be changed by passing `--prefix <path>` to the `install.sh` script.
```
rycus86: Please change to:
> The global installation folder defaults to `$HOME`, but can be changed by passing `--prefix <path>` to the `install.sh` script.

base-template.sh
```diff
@@ -45,6 +45,24 @@ are_githooks_disabled() {
...
 }
 
+#####################################################
+# Sets the ${INSTALL_DIR} variable
+
+# Returns: 0 if success, 1 otherwise
+#####################################################
+set_install_dir() {
+    INSTALL_DIR=$(git config --global githooks.installDir)
+
```
rycus86: Please change this (and also in `cli.sh`) to something like this:
```
load_install_dir() {
    INSTALL_DIR="$(git config --global githooks.installDir)"
    if [ -n "${INSTALL_DIR}" ]; then
        if [ -d "${INSTALL_DIR}" ]; then
            return
        else
            echo "! Githooks installation is corrupt! " >&2
            echo "  Install directory ${INSTALL_DIR} is missing." >&2
            echo "  Please run the Githooks install script again to fix it." >&2
        fi
    fi

    INSTALL_DIR="~/.githooks"
}
```

cli.sh
```diff
@@ -52,24 +52,44 @@ print_help_header() {
     echo "----------------------------------------------"
 }
 
+#####################################################
+# Sets the ${INSTALL_DIR} variable
+
+# Returns: 0 if success, 1 otherwise
+#####################################################
+set_install_dir() {
+    INSTALL_DIR=$(git config --global githooks.installDir)
```
rycus86: Same as in the `base-template.sh`

cli.sh
```diff
@@ -2212,6 +2232,13 @@ choose_command() {
     CMD="$1"
     [ -n "$CMD" ] && shift
 
+    if [ "$CMD" = "help" ]; then
```
rycus86: You can undo changes in the `choose_command` function.

rycus86:
It would be awesome to have some tests for these new code paths as well. :pray:

install.sh
gabyx: I set it at the end with the intention, when the installation fully worked -> its fine to set it. 
hm...

install.sh
gabyx: Fail as early as possible. Argument validation after parsing...

rycus86: I think it shouldn't fail.
If I want `./install.sh --prefix ~/some/new/path`, I'd expect that to work.

base-template.sh
gabyx: You really want to default here to "~/.githooks". Why?
I dont think this is a good idea, just very prone to undefined behavior...

install.sh
gabyx: > It would be awesome to have some tests for these new code paths as well. 🙏

Jeah for sure =)

base-template.sh
rycus86: Yes, I want to make sure that the previous behaviour is not broken, and that sensible defaults are applied for users who don't care about where things are installed.

gabyx: ok you printed a warning...

install.sh
gabyx: Ah ok, that usecase. Sure.

gabyx:
Could we mock all tests with a default install directory and test this... :-)

gabyx:
All review changes included.
Tests 108 added plus a whole run with custom installation prefix `test-installprefix.sh`

base-template.sh
rycus86: What happens when someone upgrades a single install repo?
I'm worried about this change, because someone might upgrade from an older version, not get the new config (for whatever reason), then we'd print this error message.
I think when we don't have a value set, we should silently fall back to the default value.
If the config is set, but the directory it points to doesn't exist, then we should print an error.

On a side note: are you not doing this to make Githooks working nicely in both Git Bash and WSL at the same time? Will you even have the same paths there?

tests/step-011.sh
```diff
@@ -2,6 +2,10 @@
 # Test:
 #   Direct template execution: test pre-commit hooks
 
+# Fake a default installation directory (for tests which do not install)
+mkdir -p ~/.githooks
```
rycus86: These tests need to keep working without the config being set!
When we don't have the config, they need to default to `~/.githooks`

rycus86:
My main concern is that I still don't want any error printed or failure if the install dir is not set, in that case everything should keep working as they are today, with the default path pointing to `~/.githooks`.
On testing: all existing tests should keep working without any changes, plus we need one test for the new codepaths (maybe one for install, one for base-template and one for cli).
We don't really need a whole new run with all installs going with a prefix, the existing ones should be fine for when we don't, and a handful of new ones can check the `--prefix` case.

rycus86:
On a side note: I'll be away from laptops for a few days, so I can only merge this in in the second half of next week. I'll try to still review PR changes on my mobile. Thanks!

gabyx:
For the default: I can do this for the cli.sh its already defaulting, I think for the basetemplate too, both emit a warning, which I would change to your suggestion above.

Regarding the tests: I set the configs only to ignore the warning, so with your changes I can drop these. The complete new test run was a good help to catch stupid errors, there were a hand full of these. I would leave it in. The tests have some changes with respect to detecting where it was installed. I gonna make some changes again.

gabyx:
It was far easier to make a whole new run instead of adding only some new ones, I will take a lool

gabyx:
On WSL the config is a duplikation with corrected paths

gabyx:
I made the changes you suggested :)

base-template.sh
```diff
@@ -45,6 +45,30 @@ are_githooks_disabled() {
...
+# Sets the ${INSTALL_DIR} variable
+
+# Returns: 0 if success, 1 otherwise
+#####################################################
+load_install_dir() {
+
+    INSTALL_DIR=$(git config --global --get githooks.installDir)
+
+    #shellcheck disable=SC2181
+    if [ $? -ne 0 ]; then
```
rycus86: I think a `[ -z "${INSTALL_DIR}" ]` is cleaner/better here.

base-template.sh
```diff
@@ -45,6 +45,30 @@ are_githooks_disabled() {
...
+        INSTALL_DIR=~/".githooks"
+    elif [ ! -d "$INSTALL_DIR" ]; then
+        echo "! Githooks installation is corrupt! " >&2
+        echo "  Install directory \`${INSTALL_DIR}\` is missing." >&2
+        INSTALL_DIR=~/".githooks"
+        echo "  Fallback to default directory \`${INSTALL_DIR}\`" >&2
+        echo "  Please run the Githooks install script again to fix it." >&2
+    fi
+
+    return 0
```
rycus86: Nit: you can omit this line, it doesn't change anything.

tests/step-108.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+# Test:
+#   Custom installation prefix test
+
+INSTALL_DIR="/tmp/githooks"
```
rycus86: I would avoid setting this into the same variable (or any variable to be honest) here to make sure the code doesn't actually only work because the test sets the variable that the code will also expect later.

tests/step-108.sh
```diff
@@ -0,0 +1,40 @@
...
+sh /var/lib/githooks/install.sh --prefix "$INSTALL_DIR" || exit 1
+
+if [ ! -d "$INSTALL_DIR/.githooks" ] || [ ! -f "$INSTALL_DIR/.githooks/bin/githooks" ]; then
+    echo "! Expected the install directory to be in \`$INSTALL_DIR\`"
+    exit 1
+fi
+
+if [ "$(git config --global githooks.installDir)" != "$INSTALL_DIR/.githooks" ]; then
+    echo "! Install directory in config \`$(git config --global githooks.installDir)\` "
+    echo "  is incorrect!"
```
rycus86: Keep this on the previous line, the error printed will be the last line, so better not to break it into two.

tests/step-108.sh
```diff
@@ -0,0 +1,40 @@
...
+    exit 1
+
+echo A >A.txt
+git add A.txt
+if ! git commit -a -m "Test" 2>&1 | grep -q "Githooks installation is corrupt"; then
+    echo "! Expected the installation to be corrupt (2)"
+    exit 1
+fi
+
+exit 0
```
rycus86: You can omit this line.

base-template.sh
gabyx: ah ok, wasnt sure

gabyx: isn't it so that omitting it means the exit code is the last statement hm...which anyway should not fail

tests/step-108.sh
gabyx: Lol, jeah :-)

base-template.sh
rycus86: Oh, you're probably right. Regardless, this function now can't fail anything and its exit code is not checked, so no need to `return`

gabyx:
Done.

gabyx:
I think it should be mergable now :-)

rycus86:
> I think it should be mergable now :-)
There are still some test changes to revert, please.

gabyx:
Ok, these reverts would not have been necessary... Its done.

rycus86:
Looks good, just some minor things to fix up - I'll sort them out after merging. Thanks!

PR #44, opened 2019-08-20
Small example fix.

gabyx:
Thanks for mergin.


rycus86:
This looks good, but they shouldn't bump the version number as it doesn't have any user facing changes. Can you revert all the changes except the example file?

gabyx:
Done.

PR #46, opened 2019-08-26
Missing hint text in prompt :tada:

gabyx:
Small bugfix for show_prompt. Arguments are now parsed correct.
Thx for merging :-)

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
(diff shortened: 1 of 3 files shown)


rycus86:
Thanks! Have you tested this with both GUI and without GUI dialogs?

gabyx:
Only seen it in dialog GUI mode. That its wrong. Should work also in CLI mode if the tests still parse the output enough agnositically :-)

rycus86:
Not sure there is a test for it actually, but had a look at the code, seems like it should be OK. Thanks for this!

PR #49, opened 2019-11-05
Install into bare repos.

gabyx:
Some initial fixes to allow bare repo installs.
Tests should be added...
@rycus86 Could you do this?

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,22 @@ $ sh -c "$(curl -fsSL https://raw.githubusercontent.com/rycus86/githooks/master/
 
 The GitHub URL also accepts the additional parameters mentioned above, the `https://r.viktoradam.net/githooks` URL is just a redirect to the longer GitHub address.
 
+#### Setup for bare repositories
+Because bare repositories mostly live on a server, you should setup the following:
+```shell
+cd bareRepo
+# Install Githooks into this bare repository
+git hooks install
+# Creates .githooks/trust-all marker for this bare repo
+git hooks trust
+# Automaticall accept changes to all existing and new 
+# hooks in the current repository.
+git hooks config accept trusted
+# Don't do global automatic updates, since the Githooks update
+# script should not be run in parallel on a server.
+git hooks config disable update
+```
+
 ### Templates or global hooks
 
 This script can work in one of 2 ways:
@@ -233,7 +249,7 @@ When one of them executes, it will try to find matching files in the `.githooks`
 
 #### Template folder
 
-In this approach, the install script creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`.
+In this approach, the install script creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`. For bare repositories, the hooks are installed into the `./hooks` folder on `git init --bare`.
 
 This is the recommended approach, especially if you want to selectively control which repositories use these scripts. The install script offers to search for repositories to which it will install the hooks, and any new repositories you clone will have these hooks configured.
 
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.191144-02653f
 
 #####################################################
 # Execute the current hook,
@@ -476,7 +476,7 @@ update_shared_hooks_if_appropriate() {
 
             if [ -d "$SHARED_ROOT/.git" ]; then
                 echo "* Updating shared hooks from: $SHARED_REPO"
-                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
+                PULL_OUTPUT=$(cd "$SHARED_ROOT" && env -i git pull 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Update failed, git pull output:" >&2
@@ -485,7 +485,7 @@ update_shared_hooks_if_appropriate() {
             else
                 echo "* Retrieving shared hooks from: $SHARED_REPO"
                 [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
-                CLONE_OUTPUT=$(git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
+                CLONE_OUTPUT=$(env -i git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Clone failed, git clone output:" >&2
@@ -708,10 +708,16 @@ show_prompt() {
         # else: Running fallback...
     fi
 
-    # Read from stdin
+    # Read from stdin if its available
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    # shellcheck disable=SC2229
-    read -r "$VARIABLE" </dev/tty
+    if [ -f /dev/tty ]; then
+        # shellcheck disable=SC2229
+        read -r "$VARIABLE" </dev/tty
+    else
+        echo
+    fi
+
+    # return with unset $VARIABLE ...
 }
 
 #####################################################
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.191144-02653f
+# Version: 1911.201015-bef4d7
 
 #####################################################
 # Execute the current hook,
@@ -708,16 +708,10 @@ show_prompt() {
         # else: Running fallback...
     fi
 
-    # Read from stdin if its available
+    # Read from stdin
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    if [ -f /dev/tty ]; then
-        # shellcheck disable=SC2229
-        read -r "$VARIABLE" </dev/tty
-    else
-        echo
-    fi
-
-    # return with unset $VARIABLE ...
+    # shellcheck disable=SC2229
+    read -r "$VARIABLE" </dev/tty
 }
 
 #####################################################
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.191144-02653f
+# Version: 1911.201015-bef4d7
 
 #####################################################
 # Prints the command line help for usage and
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.191144-02653f
 
 #####################################################
 # Prints the command line help for usage and
@@ -97,17 +97,26 @@ set_main_variables() {
 #####################################################
 # Checks if the current directory is
 #   a Git repository or not.
-#
+
 # Returns:
 #   0 if it is likely a Git repository,
 #   1 otherwise
 #####################################################
 is_running_in_git_repo_root() {
-    if ! git status >/dev/null 2>&1; then
-        return 1
-    fi
-
+    git rev-parse >/dev/null 2>&1 || return 1
     [ -d "${CURRENT_GIT_DIR}" ] || return 1
+}
+
+#####################################################
+# Echo if the current repository is non-bare.
+#
+# Returns: 0
+#####################################################
+echo_if_non_bare_repo() {
+    if [ "$(git rev-parse --is-bare-repository 2>/dev/null)" = "false" ]; then
+        echo "$@"
+    fi
+    return 0
 }
 
 #####################################################
@@ -401,7 +410,7 @@ git hooks trust [forget]
             touch .githooks/trust-all &&
             git config githooks.trust.all Y &&
             echo "The current repository is now trusted." &&
-            echo "  Do not forget to commit and push the trust marker!" &&
+            echo_if_non_bare_repo "  Do not forget to commit and push the trust marker!" &&
             return
 
         echo "! Failed to mark the current repository as trusted" >&2
@@ -436,7 +445,7 @@ git hooks trust [forget]
     if [ "$1" = "delete" ] || [ -f .githooks/trust-all ]; then
         rm -rf .githooks/trust-all &&
             echo "The trust marker is removed from the repository." &&
-            echo "  Do not forget to commit and push the change!" &&
+            echo_if_non_bare_repo "  Do not forget to commit and push the change!" &&
             return
 
         echo "! Failed to delete the trust marker" >&2
@@ -859,7 +868,7 @@ add_shared_hook_repo() {
         echo "# Added on $(date)" >>"$(pwd)/.githooks/.shared" &&
             echo "$SHARED_REPO_URL" >>"$(pwd)/.githooks/.shared" &&
             echo "The new shared hook repository is successfully added" &&
-            echo "  Do not forget to commit the change!" &&
+            echo_if_non_bare_repo "  Do not forget to commit the change!" &&
             return
 
         echo "! Failed to add the new shared hook repository" >&2
@@ -962,7 +971,7 @@ ${SHARED_REPO_ITEM}"
 
         echo "$NEW_LIST" >"$(pwd)/.githooks/.shared" &&
             echo "The list of shared hook repositories is successfully changed" &&
-            echo "  Do not forget to commit the change!" &&
+            echo_if_non_bare_repo "  Do not forget to commit the change!" &&
             return
 
         echo "! Failed to remove a shared hook repository" >&2
@@ -1192,7 +1201,7 @@ update_shared_hooks_in() {
 
         if [ -d "$SHARED_ROOT/.git" ]; then
             echo "* Updating shared hooks from: $SHARED_REPO"
-            PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
+            PULL_OUTPUT="$(cd "$SHARED_ROOT" && env -i git pull 2>&1)"
             # shellcheck disable=SC2181
             if [ $? -ne 0 ]; then
                 echo "! Update failed, git pull output:" >&2
@@ -1201,7 +1210,7 @@ update_shared_hooks_in() {
         else
             echo "* Retrieving shared hooks from: $SHARED_REPO"
             [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
-            CLONE_OUTPUT=$(git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
+            CLONE_OUTPUT=$(env -i git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
             # shellcheck disable=SC2181
             if [ $? -ne 0 ]; then
                 echo "! Clone failed, git clone output:" >&2
@@ -1653,7 +1662,8 @@ git hooks readme [add|update]
 
     mkdir -p "$(pwd)/.githooks" &&
         cat "$README_FILE" >"$(pwd)/.githooks/README.md" &&
-        echo "The README file is updated, do not forget to commit and push it!" ||
+        echo "The README file is updated." &&
+        echo_if_non_bare_repo "  Do not forget to commit and push it!" ||
         echo "! Failed to update the README file in the current repository" >&2
 }
 
@@ -1742,7 +1752,7 @@ git hooks ignore [trigger] [pattern...]
     done
 
     echo "The ignore file at $TARGET_DIR/.ignore is updated"
-    echo "  Do not forget to commit the changes!"
+    echo_if_non_bare_repo "  Do not forget to commit the changes!"
 }
 
 #####################################################
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.191144-02653f
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -23,7 +23,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.191144-02653f
 
 #####################################################
 # Execute the current hook,
@@ -495,7 +495,7 @@ update_shared_hooks_if_appropriate() {
 
             if [ -d "$SHARED_ROOT/.git" ]; then
                 echo "* Updating shared hooks from: $SHARED_REPO"
-                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
+                PULL_OUTPUT=$(cd "$SHARED_ROOT" && env -i git pull 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Update failed, git pull output:" >&2
@@ -504,7 +504,7 @@ update_shared_hooks_if_appropriate() {
             else
                 echo "* Retrieving shared hooks from: $SHARED_REPO"
                 [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
-                CLONE_OUTPUT=$(git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
+                CLONE_OUTPUT=$(env -i git clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Clone failed, git clone output:" >&2
@@ -727,10 +727,16 @@ show_prompt() {
         # else: Running fallback...
     fi
 
-    # Read from stdin
+    # Read from stdin if its available
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    # shellcheck disable=SC2229
-    read -r "$VARIABLE" </dev/tty
+    if [ -f /dev/tty ]; then
+        # shellcheck disable=SC2229
+        read -r "$VARIABLE" </dev/tty
+    else
+        echo
+    fi
+
+    # return with unset $VARIABLE ...
 }
 
 #####################################################
@@ -912,7 +918,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.191144-02653f
 
 #####################################################
 # Prints the command line help for usage and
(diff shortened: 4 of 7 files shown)


gabyx:
Does travis tests not run anymore?

rycus86:
There's been some issues with Travis yesterday too, it might be a bit borked.

rycus86:
Oh, here it is: https://travis-ci.org/rycus86/githooks/builds/607703891

gabyx:
I added two tests. GIT_DIR is again leaking (git version 2.22), I am curious if we should safe guard this.
Thats why I needed to do `PULL_OUTPUT=$(git --work-tree="$SHARED_ROOT" --git-dir="$SHARED_ROOT/.git" pull 2>&1)`
This sucks... I did not find which command seems to leak, sometime its arbitrarily...

gabyx:
Tests almost pass. However the whitespace test fails with the test 110. Somthing with whitespaces in shared repo url (its a local one)... but `set_shared_root` looks good... and IFS also...?

rycus86:
> Tests almost pass. However the whitespace test fails with the test 110. Somthing with whitespaces in shared repo url (its a local one)... but `set_shared_root` looks good... and IFS also...?

It's odd. Try quoting the `$(..)` expression in `execute_local_shared_hooks` and `execute_global_shared_hooks`, see if that helps? I'd expect other tests to break too though if that would be the case.

base-template.sh
```diff
@@ -471,7 +473,7 @@ update_shared_hooks_if_appropriate() {
 
             if [ -d "$SHARED_ROOT/.git" ]; then
                 echo "* Updating shared hooks from: $SHARED_REPO"
-                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
+                PULL_OUTPUT=$(git --work-tree="$SHARED_ROOT" --git-dir="$SHARED_ROOT/.git" pull 2>&1)
```
rycus86: Perhaps `PULL_OUTPUT=$(cd "$SHARED_ROOT" && git -c core.hooksPath=/dev/null pull 2>&1)` would work too?

rycus86:
```suggestion
                PULL_OUTPUT=$(cd "$SHARED_ROOT" && git -c core.hooksPath=/dev/null pull 2>&1)
```

cli.sh
```diff
@@ -1192,7 +1189,7 @@ update_shared_hooks_in() {
 
         if [ -d "$SHARED_ROOT/.git" ]; then
             echo "* Updating shared hooks from: $SHARED_REPO"
-            PULL_OUTPUT=$(cd "$SHARED_ROOT" && git pull 2>&1)
+            PULL_OUTPUT=$(git --work-tree="$SHARED_ROOT" --git-dir="$SHARED_ROOT/.git" pull 2>&1)
```
rycus86: Same:
```suggestion
            PULL_OUTPUT=$(cd "$SHARED_ROOT" && git -c core.hooksPath=/dev/null pull 2>&1)
```

install.sh
```diff
@@ -3525,6 +3525,17 @@ disable_tty_input() {
...
 
+############################################################
+# Checks whether the given directory
+#   is a Git repository (bare included) or not.
+#
+# Returns:
+#   1 if failed, 0 otherwise
+############################################################
+is_git_repo() {
+    (cd "$1" && git rev-parse --is-bare-repository >/dev/null 2>&1) || return 1
```
rycus86: I don't understand why this didn't break tests. :thinking: 
I just run `git rev-parse --is-bare-repository` on a local repository and it returns `false`, but it is a Git repo.

install.sh
```diff
@@ -3947,7 +3958,11 @@ install_into_existing_repositories() {
 
     git config --global githooks.previous.searchdir "$RAW_START_DIR"
 
-    LOCAL_REPOSITORY_LIST=$(find "$START_DIR" -type d -name .git 2>/dev/null)
+    LOCAL_REPOSITORY_LIST=$(
+        find "$START_DIR" \( -type d -and -name .git \) -or \
+            \( -type f -and -name HEAD -and -not -path "*/.git/*" \) 2>/dev/null
```
rycus86: `-type f -and -name HEAD -and -not -path "*/.git/*"` that's a bit of a weak test. Can we do better to find likely repositories but discard false positives?

install.sh
```diff
@@ -3947,7 +3958,11 @@ install_into_existing_repositories() {
...
+    )
+    echo "$START_DIR, REPOS: $LOCAL_REPOSITORY_LIST"
```
rycus86: Looks like debug statement left in?

tests/step-109.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/sh
+# Test:
+#   Set up bare repos, run the install and verify the hooks get installed
```
rycus86: Nice one!

tests/step-110.sh
```diff
@@ -0,0 +1,80 @@
...
+    echo "Neither curl nor wget is available"
+    exit 249
+fi
+
+if ! sh /var/lib/githooks/install.sh; then
+    echo "! Failed to execute the install script"
+    exit 1
+fi
+
+mkdir -p /tmp/test110/p001 &&
```
rycus86: Can you rename the `p001`, `p002` and `p003` directories to something more meaningful?
This test ended up somewhat long and it's quite hard to follow what is what.

tests/test-windows.sh
```diff
@@ -33,10 +33,12 @@ sed -i -E 's|HOOK_NAME=.*|HOOK_NAME=\${HOOK_NAME:-\$(basename "\$0")}|' "$ROOT_D
 
 if [ -n "${EXTRA_INSTALL_ARGS}" ]; then
     sed -i "s|sh \"$ROOT_DIR\"/install.sh|sh \"$ROOT_DIR\"/install.sh \${EXTRA_INSTALL_ARGS}|g" /var/lib/tests/step-* || exit 6
+    # shellcheck disable=SC1117
```
rycus86: Why is this suddenly needed?

uninstall.sh
```diff
@@ -137,10 +137,7 @@ uninstall_from_existing_repositories() {
 #   1 otherwise
 #####################################################
 is_running_in_git_repo_root() {
-    if ! git status >/dev/null 2>&1; then
-        return 1
-    fi
-
+    git rev-parse --is-bare-repository >/dev/null 2>&1 || return 1
```
rycus86: Same as above, this gives me `false` for a regular git repo.

tests/step-110.sh
```diff
@@ -0,0 +1,80 @@
+#!/bin/sh
+# Test:
+#   Trigger hooks on a bare repo with a push from a local repo.
+
+if ! curl --version && ! wget --version; then
```
rycus86: Why does this test need `curl` or `wget`?

base-template.sh
gabyx: Sure, but my changes are not due to the hooks beeing executed-> its because it needs `unset GIT_DIR` or the solution currently done.

`-c core.hooksPath=/dev/null` is certainly good

cli.sh
gabyx: here GIT_DIR is defined.
When you run `git hooks shared update` suddenly GIT_DIR is defined (perhaps from before `git hooks` invocations...

install.sh
gabyx: Really, which git version, hm.. probably not such a good idea.

install.sh
gabyx: I could not come up with a better thing. We would need a file in the main folder of the repo:
There is a `continue` in the loop below...
```
HEAD
hooks
objects
refs
```

install.sh
gabyx: Thx!

tests/step-110.sh
gabyx: Clear :-)

uninstall.sh
gabyx: Yeah, but the point is, you only want to check if its a git repo :)
This function works because if its not a git repo (bare or not)
`git rev-parse --is-bare-repository` returns not `1`

install.sh
gabyx: Yeah, but the point is, you only want to check if its a git repo :)
This function works because if its not a git repo (bare or not)
`git rev-parse --is-bare-repository` returns not `1`

gabyx: This should be faster, then `git status` which might really slow down things. and works for both bare and non-bare.

gabyx:
Some Problems in `git hooks shared add ...` "the space is gone"...

image

gabyx:
The following: `git hooks shared add "/tmp/test110/hooks" || exit 1`
results in 
image
Strange everything is quoted through the chain??

gabyx:
Whaat? Why does this split here
git hooks shared add "/tmp/test 110/hooks" || exit 1
```
choose_command() {
    echo "DEBUG: -$1- , -$2-, -$3-, -$4- "
```
image

Is this really a git alias problem???

gabyx:
Ahh fixed, damm -> The sed replacement is the issue. 
`git hooks shared add /tmp/test 110/hooks || exit 1` works

tests/test-windows.sh
gabyx: shellcheck on my machine, pre-commit hook complained

tests/step-110.sh
gabyx: Oh I dont know... it does a update I think, can I remove this, need to look what to do for the install...

gabyx:
> Added a bunch of comments, but the main gist is that I don't understand why the new Git root detection works (if I run the command on my machine, it returns `false` as expected - but then how would Githooks know it's a Git repo directory?)

> See the comments above.

rycus86:
Thanks! I'll try to have a proper look this weekend.

tests/step-110.sh
rycus86: `curl` or `wget` should only be needed if the test explicitly tests downloading something, otherwise it shouldn't be skipped just because these are missing.

uninstall.sh
rycus86: I have some issues with naming here.
It says `is_running_in_git_repo_root()` but AFAICT would work anywhere inside a git repo, not just the root.

rycus86: Also, the `git rev-parse --is-bare-repository` is still a bit interesting for me.
Can you please try if a simple `git rev-parse >/dev/null 2>&1 || return 1` would do what we want here and work for both bare repos and work trees?

tests/test-windows.sh
rycus86: Can you please check what version of shellcheck you have? Thanks!

gabyx:
OK this push should work now :-)

gabyx:
Ok, now almost all tests worked, but only the LFS test hanged -> Wiht some investigation I found, its really bad to call `git pull` and `git clone`  (git hooks update shared`) without really resetting the environment -> so the only good and robust solution is `env -i git clone` . Since a `git clone` inside `cli.sh` would run hooks `base-template.sh`-`post-checkout` eventually calling `git lfs` and leaking GIT_DIR variables and probably other variables really affect these commands.  So I guarded now these with `env -i` where happily all test work now. Strange thing...

gabyx:
The original `is_running_in_git_repo_root()` already worked everywhere inside a git repo... we can also
rename it to something else: Suggestion?

base-template.sh
```diff
@@ -703,10 +703,16 @@ show_prompt() {
         # else: Running fallback...
     fi
 
-    # Read from stdin
+    # Read from stdin if its available
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    # shellcheck disable=SC2229
-    read -r "$VARIABLE" </dev/tty
+    if [ -f /dev/tty ]; then
```
rycus86: This should be done by the test setup, interactively we expect you have TTY elsewhere. Did you leave this change in by accident during testing?

cli.sh
```diff
@@ -96,20 +96,31 @@ set_main_variables() {
...
 # Returns:
 #   0 if it is likely a Git repository,
 #   1 otherwise
 #####################################################
 is_running_in_git_repo_root() {
-    if ! git status >/dev/null 2>&1; then
-        return 1
-    fi
-
+    CURRENT_GIT_DIR_IS_BARE=$(git rev-parse --is-bare-repository)
```
rycus86: This should be just `git rev-parse || return 1` with redirecting output to devnull. We want to answer here whether we're in a git repo or not, it shouldn't care about bare or not, and should also not have side effects of setting other variables.
Not sure what's `[ -d "${CURRENT_GIT_DIR}" ] || return 1` doing right now, maybe that needs to go now?

cli.sh
```diff
@@ -96,20 +96,31 @@ set_main_variables() {
...
 
+#####################################################
+# Checks CURRENT_GIT_DIR_IS_BARE if set to `true`.
+
+# Returns:
+#   0 if it is a bare Git repository,
+#   1 otherwise
+#####################################################
+is_bare_repo() {
+    [ "$CURRENT_GIT_DIR_IS_BARE" = "false" ] || return 1
```
rycus86: This should be something like this perhaps?
```
if ! git rev-parse --is-bare-repository 2>/dev/null | grep true; then
    return 1
fi
```

rycus86: Or better yet, read it into a variable and do an equals test for it.

cli.sh
```diff
@@ -399,13 +410,18 @@ git hooks trust [forget]
...
             touch .githooks/trust-all &&
-            git config githooks.trust.all Y &&
-            echo "The current repository is now trusted." &&
-            echo "  Do not forget to commit and push the trust marker!" &&
-            return
-
-        echo "! Failed to mark the current repository as trusted" >&2
-        exit 1
+            git config githooks.trust.all Y
+        if ! mkdir -p .githooks ||
```
rycus86: Why is this change necessary?
We seem to be repeating what's already been done.

cli.sh
```diff
@@ -856,15 +876,16 @@ add_shared_hook_repo() {
...
 
-        echo "# Added on $(date)" >>"$(pwd)/.githooks/.shared" &&
-            echo "$SHARED_REPO_URL" >>"$(pwd)/.githooks/.shared" &&
-            echo "The new shared hook repository is successfully added" &&
-            echo "  Do not forget to commit the change!" &&
-            return
-
-        echo "! Failed to add the new shared hook repository" >&2
-        exit 1
+        if ! echo "# Added on $(date)" >>"$(pwd)/.githooks/.shared" ||
```
rycus86: There seem to be a few changes like this for bare repos.
I'd prefer leaving the previous print chains as-is, and perhaps add a new function: `echo_if_not_bare_repo ${msg}`
That could then suppress printing the message if not needed for bare repos.

On a side note: does printing these messages on a bare repo cause any issues?

install.sh
```diff
@@ -3947,18 +3994,24 @@ install_into_existing_repositories() {
 
     git config --global githooks.previous.searchdir "$RAW_START_DIR"
 
-    LOCAL_REPOSITORY_LIST=$(find "$START_DIR" -type d -name .git 2>/dev/null)
+    REPOSITORY_LIST=$(
```
rycus86: Why did the variable name change here by the way?

install.sh
```diff
@@ -3947,18 +3994,24 @@ install_into_existing_repositories() {
...
+        REPOSITORY_LIST=$(echo "$REPOSITORY_LIST" | sort)
     fi
 
     IFS="$IFS_NEWLINE"
-    for EXISTING in $LOCAL_REPOSITORY_LIST; do
+    for EXISTING in $REPOSITORY_LIST; do
         unset IFS
 
         EXISTING_REPO_ROOT=$(dirname "$EXISTING")
+        if ! is_git_repo "${EXISTING_REPO_ROOT}"; then
```
rycus86: This is running on the `dirname` of the repo path, how is that going to work with absolute paths?
Shouldn't this be `if ! is_git_repo "${EXISTING}"; then` ?

install.sh
```diff
@@ -3977,7 +4030,12 @@ install_into_existing_repositories() {
 ############################################################
 install_hooks_into_repo() {
     TARGET_ROOT="$1"
-    TARGET="${TARGET_ROOT}/.git"
+    TARGET="${TARGET_ROOT}"
+    IS_BARE=$(cd "${TARGET_ROOT}" && git rev-parse --is-bare-repository)
+
+    if [ "${IS_BARE}" = "false" ]; then
```
rycus86: Maybe personal preference, but this would read nicer for me:
```
if [ "${IS_BARE}" = "true" ]; then
    TARGET="${TARGET_ROOT}"
else
    TARGET="${TARGET_ROOT}/.git"
fi
```

tests/step-110.sh
```diff
@@ -0,0 +1,77 @@
...
+
+echo "Test hook from push"
+cd /tmp/test110/local || exit 1
+echo "Test" >Test
+git add Test || exit 1
+git commit -a -m "First" || exit 1
+echo "Push hook to fail"
+OUTPUT=$(git push 2>&1)
+
+#shellcheck disable=1014,2181
```
rycus86: Is it possible to resolve these issues rather than disable their checks?

uninstall.sh
```diff
@@ -137,13 +143,21 @@ uninstall_from_existing_repositories() {
 #   1 otherwise
 #####################################################
 is_running_in_git_repo_root() {
-    if ! git status >/dev/null 2>&1; then
-        return 1
-    fi
-
+    git rev-parse >/dev/null 2>&1 || return 1
```
rycus86: Shouldn't this be `is_git_repo || return 1` ?

uninstall.sh
```diff
@@ -157,7 +171,7 @@ uninstall_from_current_repository() {
         exit 1
     fi
 
-    uninstall_hooks_from_repo "$CURRENT_GIT_DIR"
+    uninstall_hooks_from_repo "$(pwd)"
```
rycus86: Why this change?

tests/step-110.sh
```diff
@@ -0,0 +1,77 @@
...
+    exit 1
+fi
+
+mkdir -p /tmp/test110/hooks &&
+    mkdir -p /tmp/test110/server &&
+    mkdir -p /tmp/test110/local || exit 1
+
+# Hooks
+cd /tmp/test110/hooks && git init || exit 1
+/var/lib/githooks/uninstall.sh --local || exit 1
```
rycus86: Why do we need to uninstall Githooks here?

rycus86:
Thanks for the changes so far! I left some more comments.

I still have concerns about this PR in general: it feels like this might be a pretty niche use-case that most people won't need, or you could only apply it to a very small number of repos where it might make sense - would you agree?

I'm thinking that maybe you'd be better off with a simple hook on bare repos without Githooks in the mix - should we explore how that might look like?

I'm trying to establish 1) what value this PR and Githooks would give you on bare repos, 2) how commonly do you need this on your bare repos, 3) is it worth the complexity?
Thanks!

gabyx:
With my changes now, which are not really much after your review I think the value is the following:
Since we would like to use Githooks company-wide, where all its update behavior come in handy, since we can manage hooks at a central place. It would be also very handy to use the exact same behavior on our repositories on the server, of course the update mechanism might be disabled and update is needed to be done manually by `git hooks update`or `git hooks shared update` since people are working in parallel (`git push`) and I dont know how that works out...
But the thing is, git has bare  and non-bare repositories, and why should Githooks not support this. Every hook manager really should, most don't, and thats cumbersome, because there is not really much of a difference considering hooks in a bare and non-bare repo. 
I wouldn't think there is now much more complexity for Githooks. The changes with `env -i` and stuff are actually strange things which really can happen also without the support for bare repos, they just came up in the test procedure, and made me puzzle and worrying about why the hack -> thinking more deeply about it goes into the direction that "are most git invocations really safe here, or might it do stupid stuff". However, we can always of course copy our own hooks to the repository of course without having githooks installed on the server ;-). The good thing is. When you make new repos on lets say Gitea, in the background the githook manager gets installed directyl from the templates and everything is setup, meaning global shared hooks are run for example and so forth :-). Its really handy I think :-). Thats why I'd think it would be valuable to have this. otherwise, we are back to managing global shared hooks manually on gitea when adding a new repo, Meaning -> People need to contact the admin -> the admin logs into the container, makes the adjustments, with githooks this is not needed anymore :-)

gabyx:
Or thats basically my motivation behind it :-). If it works out, I will see, so far, we are testing the hooks manager with a few people before we distribute a global company wide install script, which sets up everything nicely for everybody including githooks :-)

rycus86:
OK, thanks for the explanation! You're right, the latest changes don't seem much and should improve things anyway, so let's do it!

I have a few minor things left plus some unresolved comments, I'll mark those up for you to fix then it should be good to go. 🎉

Please also make sure your final commit message reflects this PR rather than "Fixes". :)

Thanks a lot for your work on this!

README.md
```diff
@@ -5,7 +5,7 @@
 
 A simple Shell script to support per-repository [Git hooks](https://git-scm.com/docs/githooks), checked into the actual repository that uses them.
 
-To make this work, it creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`. When one of them executes, it will try to find matching files in the `.githooks` directory under the project root, and invoke them one-by-one. There's more to the story though, you can read about it under the [Templates or global hooks](#Templates-or-global-hooks) section.
+To make this work, it creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone` or into `./hooks` for bare repositories. When one of them executes, it will try to find matching files in the `.githooks` directory under the project root, and invoke them one-by-one. There's more to the story though, you can read about it under the [Templates or global hooks](#Templates-or-global-hooks) section.
```
rycus86: Please revert this, here in the intro it's just confusing I think, it's enough to mention bare repos later.

README.md
```diff
@@ -118,7 +118,7 @@ You can also manage and update shared hook repositories using the [command line
 
 ## Opt-in hooks
 
-To try and make things a little bit more secure, Githooks checks if any new hooks were added we haven't run before, or if any of the existing ones have changed. When they have, it will prompt for confirmation whether you accept those changes or not, and you can also disable specific hooks to skip running them until you decide otherwise. The accepted checksums are maintained in the `.git/.githooks.checksum` file, per local repository.
+To try and make things a little bit more secure, Githooks checks if any new hooks were added we haven't run before, or if any of the existing ones have changed. When they have, it will prompt for confirmation whether you accept those changes or not, and you can also disable specific hooks to skip running them until you decide otherwise. The accepted checksums are maintained in the `.git/.githooks.checksum` file per local repository (or `./.githooks.checksum` file per bare repositories).
```
rycus86: Please revert this.

README.md
```diff
@@ -219,6 +219,19 @@ $ sh -c "$(curl -fsSL https://raw.githubusercontent.com/rycus86/githooks/master/
 
 The GitHub URL also accepts the additional parameters mentioned above, the `https://r.viktoradam.net/githooks` URL is just a redirect to the longer GitHub address.
 
+#### Setup for bare repositories
```
rycus86: This one is good

README.md
```diff
@@ -233,7 +246,7 @@ When one of them executes, it will try to find matching files in the `.githooks`
 
 #### Template folder
 
-In this approach, the install script creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`.
+In this approach, the install script creates hook templates that are installed into the `.git/hooks` folders automatically on `git init` and `git clone`. For bare repositories, the hooks are installed into the `./hooks` folder on `git init --bare`.
```
rycus86: This is also good.

base-template.sh
rycus86: Please check this one, I'm not sure this change was needed.

cli.sh
```diff
@@ -97,19 +97,28 @@ set_main_variables() {
...
 #   0 if it is likely a Git repository,
 #   1 otherwise
 #####################################################
 is_running_in_git_repo_root() {
-    if ! git status >/dev/null 2>&1; then
-        return 1
-    fi
-
+    git rev-parse >/dev/null 2>&1 || return 1
     [ -d "${CURRENT_GIT_DIR}" ] || return 1
```
rycus86: Please double-check that this still makes sense for bare repos.
The `${CURRENT_GIT_DIR}` variable is set to `git rev-parse --common-dir` or similar, no idea if that returns  something sensible for bare repos. /final

install.sh
rycus86: Please check this I think this is a bug. /final

install.sh
```diff
@@ -3977,7 +4011,13 @@ install_into_existing_repositories() {
 ############################################################
 install_hooks_into_repo() {
     TARGET_ROOT="$1"
-    TARGET="${TARGET_ROOT}/.git"
+    
+    IS_BARE=$(cd "${TARGET_ROOT}" && git rev-parse --is-bare-repository)
```
rycus86: Could you send errors to devnull? `2>/dev/null` /final

tests/step-110.sh
rycus86: What about this one? /final

uninstall.sh
```diff
@@ -121,9 +121,15 @@ uninstall_from_existing_repositories() {
...
     fi
 
-    find "$START_DIR" -type d -name .git 2>/dev/null | while IFS= read -r EXISTING; do
-        uninstall_hooks_from_repo "$EXISTING"
-    done
+    find "$START_DIR" \( -type d -and -name .git \) -or \
+        \( -type f -and -name HEAD -and -not -path "*/.git/*" \) 2>/dev/null |
+        while IFS= read -r EXISTING; do
+            EXISTING_REPO_ROOT=$(dirname "$EXISTING")
+            if ! is_git_repo "${EXISTING_REPO_ROOT}"; then
```
rycus86: Same as in the install script, this would pass in the directory name, not the path of the directory. /final

uninstall.sh
```diff
@@ -168,7 +182,15 @@ uninstall_from_current_repository() {
 #   None
 ############################################################
 uninstall_hooks_from_repo() {
-    TARGET="$1"
+    TARGET_ROOT="$1"
+
+    IS_BARE=$(cd "${TARGET_ROOT}" && git rev-parse --is-bare-repository)
```
rycus86: Please send errors to devnull here too. /final

rycus86:
Added a `/final` suffix to my remaining comments to help you search for them here.

gabyx:
Hi; =)
- Changed all with 👍 
- `EXISTING_REPO_ROOT=$(dirname "$EXISTING")` # EXISTING = "/root/bla/repo/.git"
`dirname`  returns the directory path (parent path)... so its correct.  = `/root/bla/repo`. Thats what the commands should receive...
- `if [ -f /dev/tty ]; then` is still needed to suppress error output : "file descriptor not found bla bla". Of course it works, but is just strange output.

gabyx:
Possible to merge? Thx :-)

rycus86:
Sorry, haven't had a chance to look at this yet, I'll try this week. Thanks for your patience and all the changes! :)

rycus86:
Sorry again for the delay and thanks for your patience!
One of these days I will manage to get to my laptop and review these PRs, not only on my phone...

rycus86:
Thanks! I was just looking at this, I think it's good to go.
There are a few `/final` comments you haven't replied yet, can you just double-check them amd reply before I merge?
Ignore the one about `dirname`, you explained and I realized I thought wrong what that does. :)

rycus86:
Let me know when you finished with this, then I'll get it in master. :tada:

gabyx:
Thx. I replied. Maybe check the `install_` and `uninstall_` function again. Maybe you see some errors I made...

gabyx:
FYI: This is ready.

rycus86:
Thanks a lot! I'll try to review it in an hour or so.

base-template.sh
rycus86: Out of curiosity: did you need this change for sure?
If not, I might revert, I think other places would have the same read with tty redirection.

install.sh
```diff
@@ -4028,44 +4067,53 @@ install_hooks_into_repo() {
     done
 
     # Offer to setup the intro README if running in interactive mode
-    if is_non_interactive; then
-        true # Let's skip this in non-interactive mode to avoid polluting the local repos with README files
+    # Let's skip this in non-interactive mode or in a bare repository
+    # to avoid polluting the repos with README files
+    if ! is_non_interactive && [ "${IS_BARE}" != "true" ]; then
```
rycus86: Nit: this chain is getting a bit big, it would be good to change to something like
```
if is_non_interactive; then
  true
elif [ "${IS_BARE}" != "true" ]; then
  true
else
  ...
fi
```
I'm happy to do this one after merging if you don't have time, or not at a computer.

install.sh
```diff
@@ -4028,44 +4067,53 @@ install_hooks_into_repo() {
...
+
+        # Getting the working tree (no external .git directories)
+        # see https://stackoverflow.com/a/38852055/293195
```
rycus86: Out of curiosity, why did you need this? What was wrong with the old code?

gabyx:
Made the changes, reverted the /dev/tty.

rycus86:
Thanks a lot for your hard work on this!

Issue #51, opened 2019-11-14
LFS hooks not executed when `git hooks config set disable`


gabyx:
LFS Hooks should no matter what always be executed, also if the hooks are disabled.
I make a PR.

gabyx:
Also: `execute_old_hook_if_available`
If there hooks are already lfs hooks, we execute them twice, how to avoid this???

gabyx:
May be this, is not avoidable, We just need to make sure. We correctly install `githooks` in a repo. When there are already hooks inside from LFS... hm... stupid problem...

rycus86:
Hm, is that double execution bad from LFS's perspective?
We should be able to set up a test case for this, then try and resolve the issue.
Maybe the hook execution needs to keep some state in a variable about which hook files were run already, not sure...

gabyx:
Form LFS prespective, I did not notice any issue, but its kind of bad anyway.
I think its hard to guess if the lfs hooks have already been run before any hooks by the base-template get executed... I provided #53  as a workaround.

We could also disable all LFS invocation with `sed` in the replaced hooks but then we have the problem -> these replaces hooks dont need to be `shell` or so... so we dont actually know what is in there .... sadly...

gabyx:
Thx for the merge! :+1:

PR #52, opened 2019-11-14
Bugfix: lfs hooks should always run

gabyx:
This is an urgent bugfix :-). to #51 
Thanks for squash merging.


rycus86:
Thanks for this! I expected LFS might cause some grief... :)

PR #53, opened 2019-11-14
Install option: "Purge all existing hooks" added

gabyx:
According to #51.
Dont run lfs hooks twice, installing into a repo where lfs hooks are already places is not soo good, since we are handling lfs internally.
Either
- disable any lfs invocation in the replaced hooks (cumbersome, hash?, unsafe, sed...)
- or as a workround, we provide an install option `--purge-existing` (its just handy) such that all existing hooks get deleted first.

The workaround is implemented.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.190921-99acb5
 
 #####################################################
 # Execute the current hook,
--- a/cli.sh
+++ b/cli.sh
@@ -11,7 +11,7 @@
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.181211-6b7ceb
+# Version: 1911.190921-99acb5
 
 #####################################################
 # Prints the command line help for usage and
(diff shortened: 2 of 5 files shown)


rycus86:
How about if we try to peek into the front of the hook file and find out if it's an LFS hook, in which case we just delete it rather than move it to replaced?
That would be less of an intrusive operation than just purging all existing hooks.

I believe the LFS go code does the same, reads the first kilobyte or so and checks if it looks like an LFS hook.

gabyx:
Ah, really, ok, might be a solution yes. But in that way we have more deps to LFS which is a bit sad.
I tried also. The thing is people can modify the hooks which LFS placed, and add stuff to it at the end. So you would accidentaly delete a hook the user really needed... Either sed/replace all `git lfs` invocations or leave the decision to the user, by the rahter ugly install flag I introduced.

rycus86:
Yeah, LFS wants to take over hooks, but we do too, so 🤷‍♂️🙂

rycus86:
Should we park this for now, and just add (another) workaround for LFS?
For reference, this is how LFS itself checks if it has its hooks already: https://github.com/git-lfs/git-lfs/blob/a0030968cef7b7c3c36d10f6137da98107db34c1/lfs/hook.go#L137

So, it compares the first 1024 bytes of the content only, users are free to add to the end.

gabyx:
Close this as #60 . Implements the bahavior, it works pretty nicely,

PR #56, opened 2019-11-16
Bugfix: Dont overwrite running hooks.

gabyx:
Some small bugfix. Thx for merging.

This will work for sure on UNIX. 
```shell
# Try to delete this hook first, because it could be currently running.
# The file stays around till the last file descriptor is freed.
rm -f "$TARGET_HOOK" >/dev/null 2>&1
```
On Windows it might not (cannot delete open files...) resulting in the same behavior which did not make problems so far on windows.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.141922-39ad43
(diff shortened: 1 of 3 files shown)


gabyx:
Thx, wrong stuff in the commit, now it should be mergebale. Sorry for my eagerness!

rycus86:
That's OK, thanks for the quick fix! :)

Issue #55, opened 2019-11-15
Keeping track of installed repos.


gabyx:
Scenario I am not sure about:
- RepoA (not a **single** install) triggers an update (by `git commit`) not from the console  but over a gui (no /dev/tty:
     - Uses `sh <install.sh` which launches a **global** update 👍 :
        - Update the templates and cli
        - Install into existing repos by the search dir if existing, otherwise do nothing

- Inside RepoA (not a **single** install), `git hooks update`
   - Same procedure as above...

The install into RepoA is (I think) not ensured which is a bit bad.

Wouldn't it make sense, to somehow (instead or additional to a search dir) have a list of installed repos, we keep track of, such that when ever a global update/install happens, we can install the new hooks also into the bookkeeping list (`~/.githooks/previous`) (if they are still existing at the location).

Options:
- `githooks.previous.repositories = "repoA;repoB" ` : Bookkeeping  of installed repos. or into the file `~/.githooks/previous`

```ini
[githooks "previous"]
	searchdir = ~/.bash-git-prompt
        repositories = repoA;repoB
```
What do you think. That would make update behavior a bit less cumbersome, and error prone. (user having a wrong search dir and such. Of course we can always clean the bookkeeping file or variable of any non existing repos, such that its always accurate. I tend towards a file, which is nicer to read...

gabyx:
I was thinking about that, because of update with critical bugfixes... (hopefully not much in the future)
A more reliable update procedure would be awesome :-). If I am mistaken about 
> The install into RepoA is (I think) not ensured which is a bit bad.
Do you know a test which tests this?

gabyx:
Ok we cannot update our hooks when inside RepoA's `base_template.sh` (`git commit`). 
- Maybe install an Exit Trap which checks if an install is postponing and do it there (cumbersome)
- Other solution: Make an option `githooks.failIfNotCurrentVersion` which fails any hook if the current base_template.sh has a different version than `git hooks version`. 
That would resolve running non-updated hooks, leaving the update process to the user.

The suggestion with the tracked install dirs would be nice anyway.

gabyx:
The following test fails and demonstrates the weird behavior:

```shell
#!/bin/sh
# Test:
#   Trigger update

if ! sh /var/lib/githooks/install.sh; then
    echo "! Failed to execute the install script"
    exit 1
fi

git hooks config set search-dir /tmp/test101 || exit 1

mkdir -p /tmp/test101 &&
[...]

PR #57, opened 2019-11-16
Bugfix: Never store relative search paths.

gabyx:
Small bugfix. Never store relative search paths.
This leads to undefined behavior if sudden updates happen from a hook...

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.141922-39ad43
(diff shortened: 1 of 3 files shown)


gabyx:
I just thought its a bit critical since, the search-dir is stored, togehter with the automatic update procedure which uses this search-dir alsorts of shit can happen :-) with relative paths, which is applyied directly at the `cwd` in `install.sh`. But since `realpath` is not POSIX I close this Bugfix. The #55 is in favor of this since it does not make use of the search-dir for the update mechanism. But you will still be able to update all of your repos of course, because there is a clear autoupdate registering mechanism...

PR #58, opened 2019-11-16
Bugfix: Wrong variable in LFS reinstall.

gabyx:
Samll LFS fix. Wrong variable in Reinstall. Thx for merging


gabyx:
Could catch, thx!

PR #59, opened 2019-11-16
Register repos for auto update.

gabyx:
No changes had to be made to any tests. They all work.
This is the PR for the monologue in #55. 

We improved the following:
- All repos which use Githooks hooks are now registered for autoupdates (if not single install) in `~/.gihooks/autoupdate/registered` in all circumstances, that means
   - `git init` -> first use of any hook -> registers this repo
   - `git hooks install` in repo -> register directly over `install.sh`
   - `git hooks update` in repo -> dito because of `install.sh`
   - running any Githook hook -> registering if not done (cheap)
   - running manually `install.sh` -> registering directly for any install happening...

An install triggered by a running hook now calls `DO_ONLY_UPDATE="yes" install.sh` meaning
 - if called in a single install repo, we update ony this repo.
 - in any other case, we read the `~/.gihooks/autoupdate/registered` and cycle through each entry and install the update in each registered repo. In that simple way, every repo receives the new hooks.
Add: PR #56 makes this more robust too.

I would be happy if we can get these changes in, also because it makes the whole logic better understandable. (There were some conceptual double usings for example with `--single` and an update happens.) 

Thx for the review, where I also will write some tests afterwards.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1912.161811-71df49
+# Version: 2001.241152-6f44a6
 
 #####################################################
 # Execute the current hook,
@@ -14,14 +14,15 @@
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
+    set_main_variables
+    register_for_autoupdate_if_needed
+
     if are_githooks_disabled; then
-        set_main_variables
         execute_lfs_hook_if_appropriate "$@" || return 1
         execute_old_hook_if_available "$@" || return 1
         return
     fi
 
-    set_main_variables
     export_staged_files
     check_for_updates_if_needed
     execute_lfs_hook_if_appropriate "$@" || return 1
@@ -97,6 +98,54 @@ set_main_variables() {
     # Global IFS for loops
     IFS_COMMA_NEWLINE=",
 "
+}
+
+############################################################
+# We register this repository for future potential
+# autoupdates if all of the following is true
+#   - core.hooksPath is not defined, meaning this hook
+#     needs to be in `.git/hooks`.
+#   - its not yet registered.
+#   - its a non-single install.
+#
+# Returns: None
+############################################################
+register_for_autoupdate_if_needed() {
+    if ! git config --local githooks.autoupdate.registered >/dev/null 2>&1 &&
+        [ "$(git config --local githooks.single.install)" != "yes" ] &&
+        [ ! -d "$(git config --global core.hooksPath)" ]; then
+        register_repo_for_autoupdate "$CURRENT_GIT_DIR"
+    fi
+}
+
+############################################################
+# Adds the repository to the list `autoupdate.registered`
+#  for future potential autoupdate.
+#
+# Returns: None
+############################################################
+register_repo_for_autoupdate() {
+    CURRENT_REPO="$(cd "$1" && pwd)"
+    LIST="$INSTALL_DIR/autoupdate/registered"
+
+    # Remove
+    if [ -f "$LIST" ]; then
+        TEMP_FILE=$(mktemp)
+        CURRENT_ESCAPED=$(echo "$CURRENT_REPO" | sed "s@/@\\\\\/@g")
+        sed "/$CURRENT_ESCAPED/d" "$LIST" >"$TEMP_FILE"
+        mv -f "$TEMP_FILE" "$LIST"
+    else
+        # Create folder
+        PARENT_DIR=$(dirname "$LIST")
+        if [ ! -d "$PARENT_DIR" ]; then
+            mkdir -p "$PARENT_DIR" >/dev/null 2>&1
+        fi
+    fi
+
+    # Add at the bottom
+    echo "$CURRENT_REPO" >>"$LIST"
+    # Mark this repo as registered
+    git config --local githooks.autoupdate.registered "yes"
 }
 
 #####################################################
@@ -851,11 +900,11 @@ should_run_update() {
 #####################################################
 execute_update() {
     if is_single_repo; then
-        if sh -s -- --single <"$INSTALL_SCRIPT"; then
+        if DO_UPDATE_ONLY="yes" sh -s -- --single <"$INSTALL_SCRIPT"; then
             return 0
         fi
     else
-        if sh <"$INSTALL_SCRIPT"; then
+        if DO_UPDATE_ONLY="yes" sh <"$INSTALL_SCRIPT"; then
             return 0
         fi
     fi
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 1912.161811-71df49
+# Version: 2001.241152-6f44a6
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -28,7 +28,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1912.161811-71df49
+# Version: 2001.241152-6f44a6
 
 #####################################################
 # Execute the current hook,
@@ -38,14 +38,15 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
+    set_main_variables
+    register_for_autoupdate_if_needed
+
     if are_githooks_disabled; then
-        set_main_variables
         execute_lfs_hook_if_appropriate "$@" || return 1
         execute_old_hook_if_available "$@" || return 1
         return
     fi
 
-    set_main_variables
     export_staged_files
     check_for_updates_if_needed
     execute_lfs_hook_if_appropriate "$@" || return 1
@@ -121,6 +122,54 @@ set_main_variables() {
     # Global IFS for loops
     IFS_COMMA_NEWLINE=",
 "
+}
+
+############################################################
+# We register this repository for future potential
+# autoupdates if all of the following is true
+#   - core.hooksPath is not defined, meaning this hook
+#     needs to be in `.git/hooks`.
+#   - its not yet registered.
+#   - its a non-single install.
+#
+# Returns: None
+############################################################
+register_for_autoupdate_if_needed() {
+    if ! git config --local githooks.autoupdate.registered >/dev/null 2>&1 &&
+        [ "$(git config --local githooks.single.install)" != "yes" ] &&
+        [ ! -d "$(git config --global core.hooksPath)" ]; then
+        register_repo_for_autoupdate "$CURRENT_GIT_DIR"
+    fi
+}
+
+############################################################
+# Adds the repository to the list `autoupdate.registered`
+#  for future potential autoupdate.
+#
+# Returns: None
+############################################################
+register_repo_for_autoupdate() {
+    CURRENT_REPO="$(cd "$1" && pwd)"
+    LIST="$INSTALL_DIR/autoupdate/registered"
+
+    # Remove
+    if [ -f "$LIST" ]; then
+        TEMP_FILE=$(mktemp)
+        CURRENT_ESCAPED=$(echo "$CURRENT_REPO" | sed "s@/@\\\\\/@g")
+        sed "/$CURRENT_ESCAPED/d" "$LIST" >"$TEMP_FILE"
+        mv -f "$TEMP_FILE" "$LIST"
+    else
+        # Create folder
+        PARENT_DIR=$(dirname "$LIST")
+        if [ ! -d "$PARENT_DIR" ]; then
+            mkdir -p "$PARENT_DIR" >/dev/null 2>&1
+        fi
+    fi
+
+    # Add at the bottom
+    echo "$CURRENT_REPO" >>"$LIST"
+    # Mark this repo as registered
+    git config --local githooks.autoupdate.registered "yes"
 }
 
 #####################################################
@@ -875,11 +924,11 @@ should_run_update() {
 #####################################################
 execute_update() {
     if is_single_repo; then
-        if sh -s -- --single <"$INSTALL_SCRIPT"; then
+        if DO_UPDATE_ONLY="yes" sh -s -- --single <"$INSTALL_SCRIPT"; then
             return 0
         fi
     else
-        if sh <"$INSTALL_SCRIPT"; then
+        if DO_UPDATE_ONLY="yes" sh <"$INSTALL_SCRIPT"; then
             return 0
         fi
     fi
@@ -918,7 +967,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1912.161811-71df49
+# Version: 2001.241152-6f44a6
 
 #####################################################
 # Prints the command line help for usage and
@@ -2908,9 +2957,11 @@ config_single_install() {
     fi
 
     if [ "$1" = "set" ]; then
+        git config --unset githooks.autoupdate.registered
         git config githooks.single.install yes
     elif [ "$1" = "reset" ]; then
         git config --unset githooks.single.install
+        # the repository is registered in the next hooks run
     elif [ "$1" = "print" ]; then
         if read_single_repo_information && is_single_repo; then
             echo "The current repository is marked as a single installation"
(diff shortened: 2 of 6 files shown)


gabyx:
Alright, I worked hard to make a test (:sleepy:).

install.sh
```diff
@@ -3342,23 +3395,27 @@ execute_installation() {
...
+    if ! is_update_only && setup_automatic_update_checks; then
         echo # For visual separation
     fi
 
-    # Install the hooks into existing local repositories
-    if ! should_skip_install_into_existing_repositories; then
-        if is_single_repo_install; then
-            install_hooks_into_repo "$(pwd)" || return 1
-        else
+    if is_single_repo_install; then
```
rycus86: This seems to circumvent the `--skip-install-into-existing` for a) single repos, b) existing repos on the registered list.
I'd keep the old way for this flag, if we explicitly didn't want to install into existing repos, then just don't do any of these.

install.sh
```diff
@@ -3973,6 +4043,76 @@ install_into_existing_repositories() {
...
+            fi
+
+            IFS="$IFS_NEWLINE"
+        done <"$LIST"
+
+        if [ -s "$TEMP_LIST" ]; then
+
+            echo "The following registered repositories in \`$LIST\`"
+            echo "contain also a Githooks installation and have not been updated:"
+            sed -e "s/^/ - /" <"$TEMP_LIST"
```
rycus86: I'm not sure if `sed -e` is defined on OS X (`sed -E` is I believe).
On that note though, you could avoid running `sed` here if you'd do this a few lines up:
```
                # Add to list for user
                echo " - $INSTALLED_REPO" >>"$TEMP_LIST"
```

install.sh
```diff
@@ -4076,6 +4217,39 @@ install_hooks_into_repo() {
...
+############################################################
+register_repo_for_autoupdate() {
+    CURRENT_REPO="$(cd "$1" && pwd)"
+    LIST="$INSTALL_DIR/autoupdate/registered"
+
+    # Remove
+    if [ -f "$LIST" ]; then
+        TEMP_FILE=$(mktemp)
+        CURRENT_ESCAPED=$(echo "$CURRENT_REPO" | sed "s@/@\\\/@g")
+        sed "/$CURRENT_ESCAPED/d" "$LIST" >"$TEMP_FILE"
```
rycus86: Couldn't we just do `sed "|$CURRENT_REPO|d"` and assume directories won't have the `|` character in their path?
I'm worried this double-sed might actually be more fragile.

uninstall.sh
```diff
@@ -197,6 +302,14 @@ uninstall_hooks_from_repo() {
         UNINSTALLED="yes"
     fi
 
+    # Remove all install relevant local githooks configs
+    (
```
rycus86: Why this whole subshelling? Can you make it simple, like:
```
cd "${TARGET}" && git config --local --unset githooks.single.install >/dev/null 2>&1
cd "${TARGET}" && git config --local --unset githooks.autoupdate.registered >/dev/null 2>&1
```

rycus86:
OK, added a few comments, but in general, this looks awesome! Thanks a lot!

rycus86:
Should we park this for now, and just add (another) workaround for LFS?
For reference, this is how LFS itself checks if it has its hooks already: https://github.com/git-lfs/git-lfs/blob/a0030968cef7b7c3c36d10f6137da98107db34c1/lfs/hook.go#L137

So, it compares the first 1024 bytes of the content only, users are free to add to the end.

base-template.sh
```diff
@@ -14,13 +14,14 @@
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
+    set_main_variables
+    register_for_autoupdate_if_needed
```
rycus86: Should we register this repo if Githooks is disabled here? :thinking: 
I think we should leave it as it was perhaps?

base-template.sh
```diff
@@ -98,6 +99,57 @@ set_main_variables() {
...
+#
+# Returns: 0
+############################################################
+register_for_autoupdate_if_needed() {
+    if ! git config --local githooks.autoupdate.registered >/dev/null 2>&1 &&
+        [ "$(git config --local githooks.single.install)" != "yes" ] &&
+        [ ! -d "$(git config --global core.hooksPath)" ]; then
+        register_repo_for_autoupdate "$CURRENT_GIT_DIR"
+    fi
+    return 0
```
rycus86: Nit: do we need the `return 0` in the functions where we don't actually care about the results?

base-template.sh
```diff
@@ -850,11 +902,11 @@ should_run_update() {
 #####################################################
 execute_update() {
     if is_single_repo; then
-        if sh -s -- --single <"$INSTALL_SCRIPT"; then
+        if DO_UPDATE_ONLY="yes" sh -s -- --single <"$INSTALL_SCRIPT"; then
```
rycus86: Why an environment variable, why not an argument, like the other args? (e.g. `--single --update-only`)

install.sh
```diff
@@ -3342,23 +3395,25 @@ execute_installation() {
...
     # Automatic updates
-    if setup_automatic_update_checks; then
+    if ! is_update_only && setup_automatic_update_checks; then
         echo # For visual separation
     fi
 
-    # Install the hooks into existing local repositories
     if ! should_skip_install_into_existing_repositories; then
         if is_single_repo_install; then
-            install_hooks_into_repo "$(pwd)" || return 1
```
rycus86: Why do we need to drop the early exit here?

install.sh
```diff
@@ -3342,23 +3395,25 @@ execute_installation() {
...
+            install_hooks_into_repo "$(pwd)"
         else
-            install_into_existing_repositories
+            if ! is_update_only; then
```
rycus86: Nit:
```
if is_update_only; then
  install_into_registered_repositories
else
  install_into_existing_repositories
fi
```

install.sh
```diff
@@ -3952,19 +4019,20 @@ install_into_existing_repositories() {
...
 
     IFS="$IFS_NEWLINE"
-    for EXISTING in $LOCAL_REPOSITORY_LIST; do
+    for EXISTING in $EXISTING_REPOSITORY_LIST; do
         unset IFS
 
         EXISTING_REPO_ROOT=$(dirname "$EXISTING")
-        install_hooks_into_repo "$EXISTING_REPO_ROOT"
+        install_hooks_into_repo "$EXISTING_REPO_ROOT" &&
+            register_repo_for_autoupdate "$EXISTING"
```
rycus86: Do we know at this point if `"$EXISTING"` is not a single install repository?

install.sh
```diff
@@ -3973,6 +4041,76 @@ install_into_existing_repositories() {
...
+        NEW_LIST=$(mktemp)  # the new list
+        IFS="$IFS_NEWLINE"
+        while read -r INSTALLED_REPO; do
+            unset IFS
+
+            if [ ! -d "$INSTALLED_REPO" ] ||
+                echo "$EXISTING_REPOSITORY_LIST" | grep -q "$INSTALLED_REPO"; then
+                # Add to new list
+                echo "$INSTALLED_REPO" >>"$NEW_LIST"
+            elif (cd "$INSTALLED_REPO" && "$(git config --local githooks.single.install)" = "yes"); then
```
rycus86: Is this a test like this in Shell? Does it not need the `[ ]`?

install.sh
```diff
@@ -3973,6 +4041,76 @@ install_into_existing_repositories() {
...
+
+        if [ -s "$TEMP_LIST" ]; then
+
+            echo "The following registered repositories in \`$LIST\`"
+            echo "contain also a Githooks installation and have not been updated:"
+            sed -E "s/^/ - /" <"$TEMP_LIST"
+            printf 'Do you want to install to all of them? [Yn] '
+
+            read -r DO_INSTALL </dev/tty
+            if [ "$DO_INSTALL" = "n" ]; then
```
rycus86: Please also allow `N` here.

install.sh
```diff
@@ -3973,6 +4041,76 @@ install_into_existing_repositories() {
...
+        # - non-existing repos
+        # - already installed
+        # - if marked as single install.
+        TEMP_LIST=$(mktemp) # temporary list for user
+        NEW_LIST=$(mktemp)  # the new list
+        IFS="$IFS_NEWLINE"
+        while read -r INSTALLED_REPO; do
+            unset IFS
+
+            if [ ! -d "$INSTALLED_REPO" ] ||
```
rycus86: Sorry, I can't really work it out from this.
What is this `[ ! -d "$INSTALLED_REPO" ] || echo "$EXISTING_REPOSITORY_LIST" | grep -q "$INSTALLED_REPO";` case?
We had it in the `autoupdate/registered` file, but the repo doesn't exist or if the user let us search for the directories and it was on it? (don't understand the meaning of this)

Also can you explain what are `$LIST`, `$NEW_LIST` and `$TEMP_LIST` please? Hopefully we can work out a better name for them.

install.sh
```diff
@@ -3973,6 +4041,76 @@ install_into_existing_repositories() {
...
+                # Add to list for user
+                echo "$INSTALLED_REPO" >>"$TEMP_LIST"
+            fi
+
+            IFS="$IFS_NEWLINE"
+        done <"$LIST"
+
+        if [ -s "$TEMP_LIST" ]; then
+
+            echo "The following registered repositories in \`$LIST\`"
```
rycus86: How have they not been updated if they were on the `autoupdate/registered` list?
Did the user have to make some manual changes for this case somewhere?

install.sh
```diff
@@ -3987,6 +4125,7 @@ install_hooks_into_repo() {
     if [ ! -w "${TARGET}/hooks" ]; then
         # Try to create the .git/hooks folder
         if ! mkdir "${TARGET}/hooks" 2>/dev/null; then
+            echo "! Could not install into \`$TARGET\` because there is not write access."
```
rycus86: Nit:
```suggestion
            echo "! Could not install into \`$TARGET\` because there is no write access"
```

rycus86:
Added some comments after another review. :+1: 
This also needs rebasing on master.
Thanks!

gabyx:
Rebased to master and included your review + better comments etc.

gabyx:
Could you give it another review. I will check the tests tonight.

rycus86:
Hey! Thanks, I'll try to review when I'm at a computer later. From a quick glance, it looks like there are lots of unresolved comments from last time, could you double check them and reply with your thoughts on them? Thanks a lot! (and happy new year 🎉)

install.sh
```diff
@@ -38,14 +38,15 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
+    set_main_variables
+    register_for_autoupdate_if_needed
```
gabyx: We always register (meaning this repo might be prone to updates, which is good). 
No matter if the hooks are disabled.

install.sh
gabyx: Its ok as it is IMO. We want both to happen when we are not int update mode (user requested an update, not triggered from the hooks)

install.sh
```diff
@@ -4094,25 +4162,27 @@ install_into_existing_repositories() {
 
     git config --global githooks.previous.searchdir "$RAW_START_DIR"
 
-    LOCAL_REPOSITORY_LIST=$(
+    EXISTING_REPOSITORY_LIST=$(
```
gabyx: This variable is used latter when `installing_into_register`

gabyx:
I rebased and revised all old review marks (some did not even apply anymore).
Furthermore I squashed into one commit.

gabyx:
Test should now work.

rycus86:
Jad a look over this, looks all fine to me, thanks a lot!
I'll do a final pass when I'm at my computer then it should be good to go in.
I have some minor nits, but I can fix them up later on master if they still bother me at that point. :)

Thanks again for your hard work on this!

gabyx:
Thx. and happy new year :-)

rycus86:
These all look good to me and I almost merged this PR, but then I noticed the new test is actually failing on Travis: https://travis-ci.org/rycus86/githooks/pull_requests

Can you please work out why that is? Thanks!

gabyx:
Sure, sorry, I thought I fixed it already, probably a typo. Fix in 2-3hours...

gabyx:
Ok the mistake was a `</dev/tty` in uninstall.sh.
Actually do you know why we actually use `/dev/tty` instead of stdin in `install.sh`. I have not yet got that part...

gabyx:
Almost done. Some wrong `sed`
`CURRENT_ESCAPED=$(echo "$CURRENT_REPO" | sed "s@/@\\\\\/@g")`
instead of
`CURRENT_ESCAPED=$(echo "$CURRENT_REPO" | sed "s@/@\\\/@g")`

gabyx:
And forgot to skip last update test in 116 if curl/wget is not there
```
if ! curl --version && ! wget --version; then
    # We skip the update for neither curl or wget
    exit 0
fi
```

gabyx:
Now they all shall pass... :-P

gabyx:
Do you understand why **all** tests work except the coverage with test116 (my added one?)

rycus86:
Ah, that one is a special little snowflake... To get coverage on it, we need the otherwise POSIX scripts to run with Bash, and there are some limitations of `kcov` to work around.
Have a look at https://github.com/rycus86/githooks/blob/master/tests/coverage.sh for the things we replace there. My best guess would be perhaps the `s/sh/bash/` rewrite doesn't like the `DO_UPDATE=yes sh ...` or something, but it's just a guess.

I'd appreciate it if you'd try to fix it (you can run it locally with `sh tests/coverage.sh` or just that one step with `sh tests/coverage step-116` I think), but if you can't work it out, I'll merge the PR as it look OK otherwise, then will try to work out the coverage issue if I'll ever have time.

Thanks!

tests/test-whitespace.sh
```diff
@@ -9,7 +9,7 @@ EOF
 
 export ADDITIONAL_INSTALL_STEPS='
 # add a space in paths and wrap in double-quotes
-RUN find /var/lib/tests -name "*.sh" -exec sed -i -E "s|/tmp/test([0-9]+)|\"/tmp/test \1\"|g" {} \;
+RUN find /var/lib/tests -name "*.sh" -exec sed -i -E "s|/tmp/test([0-9.]+)|\"/tmp/test \1\"|g" {} \;
```
rycus86: This shouldn't be necessary, `/tmp/test109 .1/something` should be a valid path.
On that note, why 109 and not 116 ? :)

gabyx:
This regex needed to also match /tmp/test109.1
thats why I added the „.“ 

I will change to 116.1 of course :). The coverage. I ll have a look tomorrow. 

Von meinem iPhone gesendet

gabyx:
- I could not find any wrong replaced stuff in all scripts in the container of `coverage.sh` with
```
r=$(pwd)
docker run -v "$r:/repo" -it githooks:coverage bash -c "cp -r /var/lib/githooks /repo"
```
- I tested the test 116 with bash in `test-alpine-curl` it seems to work.

Hm...

gabyx:
So far, I could not really find out what the problem is. 
It seems that kcov, seems to hang on the command 
```
echo 'Y
/tmp
' | sh /var/lib/githooks/install.sh || exit 1
```
only 
```
sh /var/lib/githooks/install.sh || exit 1
```
Works, do you know why?

Can we merge this, and look for the cause later?

rycus86:
Yep, thanks for this!
Will try to work out what's wrong with the coverage. It's a bit fiddly, I usually just run coverage on a single test, then from the covered lines I can see how far the execution gets, then make a guess what's wrong with the next line. 🤷‍♂️

rycus86:
@gabyx I just updated and got a bunch of messages, saying:
```
fatal: --local can only be used inside a git repository
```

Have you seen this by any chance? I suspect it might be related to this change.

gabyx:
Ah jep, wuii, this is related to my changes:
When we search for repos, I found also
`cd /home/gabyx/Desktop/Repository/BUILD/GRSFramework/CMakeFiles/git-data/HEAD`
which is not a git repo. I think we should safe guard after the find with a check if we are inside a git repo if not, we exclude it. 
I make a fix.

gabyx:
May be something like that:
```shell
REPOSITORY_LIST=$(
        find "$START_DIR" \( -type d -and -name .git \) -or \
            \( -type f -and -name HEAD -and -not -path "*/.git/*" \) 2>/dev/null
    )

    EXISTING_REPOSITORY_LIST=""
    echo "$REPOSITORY_LIST" | while IFS= read -r EXISTING; do
        if [ -f "$EXISTING" ]; then
            # Strip HEAD file
            EXISTING=$(dirname "$EXISTING")
        fi
        if is_git_repo "$EXISTING"; then 
            EXISTING_REPOSITORY_LIST=$(echo -e '$EXISTING\n$EXISTING_REPOSITORY_LIST')
        fi
    done
```

gabyx:
See #76

rycus86:
@gabyx not sure if it's related to this PR, but I've noticed with the last 2 updates on my work machine that the installer creates `hooks` folders in the repositories it was run in, at their root folder, not under their `.git` subfolder as I'd expect. Can you check if the same happens to you too? When prompted to update, I just say yes search my main dev area, then say yes to install it into previously registered repos yoo, but skip creating README files.

Thanks!

gabyx:
I think this is the artefact I tried to correct in #77. For me this does not happen anymore on @master.
Maybe delete the autoupdate.register file?
Does it happen with a new install? I mean the tests all work, how could that be?
Strange.

rycus86:
Ah yes, that could be it. Maybe there's a way to automatically discard those invalid entries?

gabyx:
I though I already do that in 
https://github.com/rycus86/githooks/blob/dd8d92cf7398d855aba6aad3b05d5523f448f905/install.sh#L4301
Aha, but I dont check if its really a git-dir. I only check if its a git repo. I make a PR

gabyx:
See #79

gabyx:
😰 ;-)

PR #66, opened 2019-11-21
Bugfix core.hooksPath

gabyx:
There are some really unseen errors in the changes I made. Please have look.
Its not obvious. But the path I took, now makes sense.
See the annotated comments.

Honestly, not to be unfair, but there are a lot of comments in the code base which are trivially unecessary and only make the code much harder to read:
Such things... Would be good to clean those once...
```shell
    # Delete the hook templates
    remove_existing_hook_templates "$TARGET_TEMPLATE_DIR"

    # Uninstall all shared hooks
    uninstall_shared_hooks

    # Uninstall all cli
    uninstall_cli
```
My point: If you choose your function name wisely, there is no need for a comment, which is almost the case everywhere :-)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 1911.212020-f03ac5
+# Version: 1911.212022-20c6fe
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -23,7 +23,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.212020-f03ac5
+# Version: 1911.212022-20c6fe
 
 #####################################################
 # Execute the current hook,
@@ -912,7 +912,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.212020-f03ac5
+# Version: 1911.212022-20c6fe
 
 #####################################################
 # Prints the command line help for usage and
@@ -3750,7 +3750,7 @@ search_pre_commit_sample_file() {
         read -r ACCEPT </dev/tty
 
         if [ "$ACCEPT" = "y" ] || [ "$ACCEPT" = "Y" ]; then
-            TARGET_TEMPLATE_DIR=$(dirname "$HIT")
+            TARGET_TEMPLATE_DIR="$HIT"
             return
         fi
         IFS="$IFS_NEWLINE"
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 1911.201542-e35c79
+# Version: 1911.212020-f03ac5
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -23,7 +23,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.201542-e35c79
+# Version: 1911.212020-f03ac5
 
 #####################################################
 # Execute the current hook,
@@ -912,7 +912,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.201542-e35c79
+# Version: 1911.212020-f03ac5
 
 #####################################################
 # Prints the command line help for usage and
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -3333,25 +3333,18 @@ execute_installation() {
         disable_tty_input
     fi
 
-    # Find the directory to install to
     if is_single_repo_install; then
         ensure_running_in_git_repo || return 1
         mark_as_single_install_repo
     else
         prepare_target_template_directory || return 1
-    fi
```
gabyx: Combine this together, less ifs => better code

install.sh
```diff
@@ -3333,25 +3333,18 @@ execute_installation() {
...
-
-    # Install the hook templates if needed
-    if ! is_single_repo_install; then
         setup_hook_templates || return 1
         echo # For visual separation
     fi
 
-    # Install the command line helper tool
```
gabyx: [doc] We can see this... :-P

install.sh
```diff
@@ -3431,12 +3426,6 @@ parse_command_line_arguments() {
             NON_INTERACTIVE="yes"
         elif [ "$p" = "--single" ]; then
             SINGLE_REPO_INSTALL="yes"
-
```
gabyx: Moved to the bottom

install.sh
```diff
@@ -3459,15 +3448,17 @@ parse_command_line_arguments() {
             USE_CORE_HOOKSPATH="yes"
             # No point in installing into existing when using core.hooksPath
             SKIP_INSTALL_INTO_EXISTING="yes"
-
```
gabyx: Moved to the bottom

install.sh
```diff
@@ -3459,15 +3448,17 @@ parse_command_line_arguments() {
...
-            # Using core.hooksPath implies it applies to all repo's
-            if [ "$SINGLE_REPO_INSTALL" = "yes" ]; then
-                echo "! Cannot use --single and --use-core-hookspath together" >&2
-                exit 1
-            fi
+        elif [ "$prev_p" = "--use-core-hookspath" ] && (echo "$p" | grep -qvE '^\-\-.*'); then
```
gabyx: Forgotton to parse this...

install.sh
```diff
@@ -3459,15 +3448,17 @@ parse_command_line_arguments() {
...
+            TARGET_TEMPLATE_DIR="$p"
         fi
         prev_p="$p"
     done
+
```
gabyx: Do the validation here, better.

install.sh
```diff
@@ -3599,9 +3590,7 @@ prepare_target_template_directory() {
         return 1
     fi
 
-    if [ "$USE_CORE_HOOKSPATH" = "yes" ]; then
-        set_githooks_directory "$TARGET_TEMPLATE_DIR"
-    fi
+    set_githooks_directory "$TARGET_TEMPLATE_DIR"
```
gabyx: Argh... this function made me go nuts...
Here thats the only single call to it, which makes understanding the code better.
First we find the template dir `$TARGET_TEMPLATE_DIR`:
- if  `--use-core-hookspath`  is set -> it will exist
-  if `--use-core-hookspath` not set -> its of the form `<dir>/hooks`, where <dir> exists, or  only `<dir>`

install.sh
```diff
@@ -3649,14 +3638,8 @@ find_git_hook_templates() {
             read -r MARK_AS_TEMPLATES </dev/tty
 
             if [ "$MARK_AS_TEMPLATES" = "y" ] || [ "$MARK_AS_TEMPLATES" = "Y" ]; then
-                TEMPLATE_DIR=$(dirname "$TARGET_TEMPLATE_DIR")
```
gabyx: Thats clearly a bug, $TARGE_TEMPLATE_DIR contains `<dir>/hooks` and here we drop the `hooks` part.

install.sh
```diff
@@ -3666,7 +3649,6 @@ find_git_hook_templates() {
 
     if [ "${SETUP_NEW_FOLDER}" = "y" ] || [ "${SETUP_NEW_FOLDER}" = "Y" ]; then
         setup_new_templates_folder
-        if [ "$TARGET_TEMPLATE_DIR" != "" ]; then return; fi
```
gabyx: Not needed... make it simple

install.sh
```diff
@@ -3683,19 +3665,15 @@ mark_directory_as_target() {
         return
     fi
 
-    if [ "$2" != "" ]; then
```
gabyx: Thats a but, here we add the `$2` below we  check for writable -> not according to the doc, its strange behavior.
Here **only* the first dir is checked, Thats so far what I could made up by looking at the calls, exactly what should have been achieved.... These caveats came in related to the coreHooksPath code paths i think....

uninstall.sh
```diff
@@ -404,21 +413,10 @@ uninstall() {
     git config --global --unset githooks.previous.searchdir
     git config --global --unset githooks.disable
     git config --global --unset githooks.installDir
+    git config --global --unset githooks.pathForUseCoreHooksPath
+    git config --global --unset githooks.useCoreHooksPath
     git config --global --unset alias.hooks
 
-    if using_hooks_path; then
```
gabyx: No ifs, rather prefere to uninstall all stuff always (its an uninstall, try not to leave stuff behind -> which is not guaranteed too, but that needs fixes as well, maybe somewhen else), moved the code here to the top, and the unset of configs together with the others

gabyx:
Please ignore this PR for now. It needs workd

PR #67, opened 2019-11-21
Bugfix: core hookspath

gabyx:
The path to `--use-core-hookspath` is not parsed and set.
This is a small bug fix, with some minor cosmetics. 

- Less `ifs`, less early returns (never good to understand)
- Some errors in function `mark_directory_as_target`.
 
P.S:
Where I am not sure, but nothing to this PR. After step `# 4. Setup new folder if running` in `find_git_hook_templates` I am not sure what happens if we are there and `--use-core-hookspath`. I think its fine. Its a bit confusing... But in another way its also clear you choose a template dir and then it gets used as the core.hooksPath...

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 1911.260018-b70412
+# Version: 1911.252357-164a99
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -28,7 +28,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1911.260018-b70412
+# Version: 1911.252357-164a99
 
 #####################################################
 # Execute the current hook,
@@ -917,7 +917,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 1911.260018-b70412
+# Version: 1911.252357-164a99
 
 #####################################################
 # Prints the command line help for usage and
@@ -3438,10 +3438,6 @@ execute_installation() {
         mark_as_single_install_repo
     else
         prepare_target_template_directory || return 1
-    fi
-
-    # Install the hook templates if needed
-    if ! is_single_repo_install; then
         setup_hook_templates || return 1
         echo # For visual separation
     fi
@@ -3518,6 +3514,8 @@ load_install_dir() {
 # Sets ${SKIP_INSTALL_INTO_EXISTING} for --skip-install-into-existing
 # Sets ${INSTALL_DIR} for --prefix
 # Sets ${TARGET_TEMPLATE_DIR} for --template-dir
+# Sets ${INSTALL_ONLY_SERVER_HOOKS} for --only-server-hooks
+# Sets ${USE_CORE_HOOKSPATH} for --use-core-hookspath
 #
 # Returns: None
 ############################################################
@@ -3690,14 +3688,33 @@ mark_as_single_install_repo() {
 ############################################################
 prepare_target_template_directory() {
     if [ -z "$TARGET_TEMPLATE_DIR" ]; then
-        find_git_hook_templates
+        # Automatically find a template directory
+        if ! find_git_hook_templates; then
+            echo "Git hook templates directory not found" >&2
+            return 1
+        fi
+    else
+        # The user provided a template directory, check it and
+        # add `hooks` which is needed.
+        if [ ! -d "$TARGET_TEMPLATE_DIR" ]; then
+            echo "Git hook templates directory does not exists" >&2
+            return 1
+        else
+            TARGET_TEMPLATE_DIR="$TARGET_TEMPLATE_DIR/hooks"
+        fi
     fi
 
+    # TARGET_TEMPLATE_DIR is now `<template-dir>/hooks`
+    # Create the `hooks` directory if it does not yet exist:
     if [ ! -d "$TARGET_TEMPLATE_DIR" ]; then
-        echo "Git hook templates directory not found" >&2
-        return 1
+        if ! mkdir -p "$TARGET_TEMPLATE_DIR" >/dev/null 2>&1; then
+            echo "Could not create folder \`$TARGET_TEMPLATE_DIR\`"
+            return 1
+        fi
     fi
 
+    # Up to now the directories would not have been set if
+    # --use-core-hookspath is used, we set it now here.
     if [ "$USE_CORE_HOOKSPATH" = "yes" ]; then
         set_githooks_directory "$TARGET_TEMPLATE_DIR"
     fi
@@ -3715,7 +3732,7 @@ prepare_target_template_directory() {
 find_git_hook_templates() {
     # 1. from environment variables
     mark_directory_as_target "$GIT_TEMPLATE_DIR" "hooks"
-    if [ "$TARGET_TEMPLATE_DIR" != "" ]; then return; fi
+    if [ "$TARGET_TEMPLATE_DIR" != "" ]; then return 0; fi
 
     # 2. from git config
     if [ "$USE_CORE_HOOKSPATH" = "yes" ]; then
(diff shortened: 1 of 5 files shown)


install.sh
```diff
@@ -3679,24 +3672,22 @@ find_git_hook_templates() {
 ############################################################
 mark_directory_as_target() {
     TARGET="$1"
-    if [ "$TARGET" = "" ]; then
+    if [ -z "$TARGET" ]; then
         return
     fi
 
-    if [ "$2" != "" ]; then
-        TARGET="${TARGET}/$2"
```
gabyx: Already settings this here, make the next if, not the thing the function actually wants to test! -> I rewrote, for clearer appearance

uninstall.sh
```diff
@@ -404,21 +413,10 @@ uninstall() {
     git config --global --unset githooks.previous.searchdir
     git config --global --unset githooks.disable
     git config --global --unset githooks.installDir
+    git config --global --unset githooks.pathForUseCoreHooksPath
```
gabyx: Uninstall everything, please... :-)

gabyx:
Lets test.

gabyx:
All test pass.

rycus86:
I'll check this PR later today. Thanks!

rycus86:
Hm, I'm a bit confused about this PR, what is the fix actually in here?
I see that you rewrote a bunch of ifs and return statements (most of which I disagree with :) ), changed some braces around variable names, but the logic doesn't seem to change.

If there is a bug, could you please have the PR fix that only?
If you'd like to discuss those code style/program flow changes, I'm open to it, but let's talk about them on an issue maybe first (maybe a PR is better to comment on the code itself), but the important thing is that we make that issue/PR about the styling changes, not the bugfix - hope that sounds OK?

gabyx:
Clear. Was a bit frustrated, with bash and IMO strange code flow latetly ...
**I comitted only the relevant stuff.**

The bug fixes is in `install.sh`
- L3686, which is not right. (according to the function what it should do)
- `$ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --use-core-hookspath /home/public/.githooks` does not work -> Fixed it. L3501
- The control flow in `find_git_hook_templates` has some errors after step 4 when `--use-core-hookspath` is used and we reach to step 4 in this function. Think about it. My fix is what should be done to set the correct paths, however my fix doesnt make the control flow better.
- Added a another warning on L4305. If the user messes up (like me :-P)

Feel free to incorporate what you think is valuable 
I came across this when trying to experiment with install.sh in a script on the server... 
And I had also some mess with the state and such...

rycus86:
Sure, no probs and thanks for the changes!
I'll have another look at this shortly.

install.sh
```diff
@@ -3791,7 +3796,7 @@ search_pre_commit_sample_file() {
     for HIT in $(find "$START_DIR" -path "*templates/hooks/pre-commit.sample" 2>/dev/null); do
         unset IFS
 
-        HIT=$(dirname "$HIT")
+        HIT=$(dirname "$(dirname "$HIT")")
```
gabyx: Lets set what the function says, search the template folder, its not with an added `./hooks`.

gabyx:
Ok, now its correct.

gabyx:
Wua, thats hard to grasp controlflow, damm, not its better. :-|

install.sh
```diff
@@ -3838,7 +3842,9 @@ setup_new_templates_folder() {
     if ! is_dry_run; then
         if mkdir -p "${TILDE_REPLACED}/hooks"; then
             # Let this one go with or without a tilde
-            set_githooks_directory "$USER_TEMPLATES"
+            if [ -z "$USE_CORE_HOOKSPATH" ]; then
```
gabyx: Guard! -> Set it at the end...

install.sh
```diff
@@ -3642,6 +3646,8 @@ prepare_target_template_directory() {
         return 1
     fi
 
+    # Up to now the directories would not have been set if
```
gabyx: Now set it.

install.sh
```diff
@@ -3692,10 +3698,11 @@ find_git_hook_templates() {
             read -r MARK_AS_TEMPLATES </dev/tty
 
             if [ "$MARK_AS_TEMPLATES" = "y" ] || [ "$MARK_AS_TEMPLATES" = "Y" ]; then
-                TEMPLATE_DIR=$(dirname "$TARGET_TEMPLATE_DIR")
-
-                if ! set_githooks_directory "$TEMPLATE_DIR"; then
-                    echo "! Failed to set it up as Git template directory" >&2
+                if [ -z "$USE_CORE_HOOKSPATH" ]; then
```
gabyx: Guard! Set it at the end!

gabyx:
@rycus86 : I added a test 112 which is simple. It does currently not work on master.
Are you not also the opinion, that my changes (not much) might go into the right direction?
At least it fixes the test. Also I checked again:
Reaching step 4 with `--use-core-hookspath` is also perfectly fine.
`# 4. Setup new folder if running interactively and no folder is found by now`.
May be you have some time to have a look.

tests/step-112.sh
```diff
@@ -0,0 +1,40 @@
...
+# check if hooks are inside the template folder.
+for hook in pre-push pre-receive pre-commit; do
+    if ! [ -f "$templateDir/hooks/$hook" ]; then
+        echo "! Hooks were not installed successfully"
+        exit 1
+    fi
+done
+#shellcheck disable=2012
+count="$(ls "$templateDir/hooks/" | wc -l)"
+if [ "$count" != "19" ]; then
```
rycus86: Aren't there only 7 server hooks?

tests/step-112.sh
```diff
@@ -0,0 +1,40 @@
...
+
+mkdir -p /tmp/test112
+cd /tmp/test112 && git init --bare || exit 1
+
+if grep -r 'github.com/rycus86/githooks' /tmp/test112/; then
+    echo "! Hooks were installed ahead of time"
+    exit 1
+fi
+
+mkdir -p ~/.githooks/templates
```
rycus86: If you do `mkdir -p ~/.githooks/templates/hooks` here, then this test already works on `master`.

rycus86:
Thanks for the test!
I left some comments on it, that it looks like it's more a test setup issue rather than an install script issue.
Still, I don't quite get why are there `corePath` related changes in this PR if the problem is with the template directory? (if I understand correctly)

rycus86:
For this one

> $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --use-core-hookspath /home/public/.githooks does not work -> Fixed it. L3501

I see what you mean now.
It feels like a much cheaper fix would be:
- `git config --global core.hooksPath $argFromShell`
- `do the existing logic`

install.sh
```diff
@@ -4304,15 +4310,26 @@ set_githooks_directory() {
         git config --global githooks.useCoreHooksPath yes
         git config --global githooks.pathForUseCoreHooksPath "$1"
         git config --global core.hooksPath "$1"
+
+        CURRENT_CORE_TEMPLATE_DIR=$(git config --global init.templateDir)
+        if [ -n "$(ls -A "$CURRENT_CORE_TEMPLATE_DIR/hooks" 2>/dev/null)" ]; then
+            echo "! The \`core.templateDir\` setting is currently set to \`$CURRENT_CORE_TEMPLATE_DIR\`" >&2
```
rycus86: `s/core.templateDir/init.templateDir/`

tests/step-112.sh
gabyx: I know, but is this the behavior one wants?

rycus86: Yeah, I'm not sure.
Perhaps we should create that directory if `init.templateDir` is set?

install.sh
```diff
@@ -3872,6 +3874,10 @@ setup_hook_templates() {
         HOOK_NAMES="$MANAGED_HOOK_NAMES"
     fi
 
+    # The template target directory exists,
+    # make sure the added paths (e.g. `hooks`) do too
+    mkdir -p "$TARGET_TEMPLATE_DIR" >/dev/null 2>&1
+
```
gabyx: Thats where I create the directory. The comment should be: "The template target directory partially exists, make sure the added parts (e.g. hooks) do too...

rycus86:
What do you meant by this, sorry? (I think line numbers have got messed up)
> L3686, which is not right. (according to the function what it should do)

rycus86:
Please check out https://github.com/rycus86/githooks/pull/69 , I think this is not actually an issue, but the README was quite misleading on the parameter usage.

gabyx:
One question: When a valid existing `init.tempalteDir` is set but no `hooks` subdirectory. A normal installation should work with out specifying `--template-dir`. (shouldn't it?)

This is my fix. I will update accordint to the PR #69 .

rycus86:
In that case, you template dir might not have been valid, and I think the installer will continue trying the default location in `/usr/...`, then offer to search for it, then offer to create one (by default at `~/.githooks/templates` (no `hooks` subfolder in this one).

Sorry about the back and forth in this PR, I couldn't remember these things until I started writing those tests last night, then ot clicked that it should be working, only the README is wrong.

tests/step-112.sh
gabyx: Sure, but this test had nothing to do with server hooks... :-) Just demonstrating the failure..

tests/step-112.sh
gabyx: Yeah, definitely. Thas included in my fix. Wait I prepare the rebase changes...

tests/step-112.sh
gabyx: Ups, jeah you are right, I set up a bare repo, why did it install not 7 hooks -> because I did not explicitly state on the command line `--only-server-hooks` during install.sh. So all is fine

gabyx:
Closed in favor of a new one, to restart again.

PR #68, opened 2019-11-21
Fix test 111 

gabyx:
Dito

Why does it need the `-E` ?


gabyx:
Thx!

gabyx:
git is just f**** great ;-)

rycus86:
Ah yes. On some OSes, `grep` defaults to regex mode, but not on the `alpine` images apparently.

rycus86:
Thanks a lot for fixing!

PR #69, opened 2019-11-25
Template dir fixes :green_heart:

rycus86:
Related to #67 

I believe this was just a documentation mistake.
We don't really want to create directories anywhere, so we get implicit consent from the user by:
- getting the path they want from `--template-dir`, and
- requiring that the target template directory exists that we've been given

This PR fixes the README, and adds two tests to show the install script invocation.


gabyx:
Ok, that was my first thought, at the beginning, but was a bit confused. I would merge this!

gabyx:
I will then update my PR to master and simplify it!

Issue #70, opened 2019-11-25
Template Dir Discussion and Improvements


gabyx:
I looked again over the tests 113/114.
I really don't understand what is tested in 113 especially (114 is strange too) ?

```shell
# run the install, and select installing hooks into existing repos
echo 'n
y
/tmp/test113
' | sh /var/lib/githooks/install.sh --template-dir ~/.githooks/templates || exit 3

# check if hooks are inside the template folder.
if ! sh /var/lib/githooks/cli.sh list | grep test-hook; then
    echo "! Hooks were not installed successfully"
    exit 4
fi
```
First, I dont understand why `git hooks list` would be a sufficient test if the hooks are placed in `~/.githooks/templates`. I think its not...

Second, when you add 
```
ls -al ~/.githooks/templates
exit 1
```
you see that all hooks are directly installed into `~/.githooks/templates`. and no **subfolder** `hooks` is created. There is some missconception that `--template-dir` is directly the path where the hooks are installed, that should not be, since `init.templateDir` is also not in this manner.
the `--template-dir` path should be the same as `init.templateDir` meaning its the **template directory** as git describes. Any other redefinition for `--template-dir` does not make sense and is only hard to understand for the user.

There is a little mess with the install folder, when to create/when not, and when to add `hooks` and when not. All that I tried to address in #71 , where I cleaned the control paths such that it avoids bugs and unforeseen behavior. Can you please have a look at it again in detail.

rycus86:
First I couldn't remember either, but then I worked it out by following the existing logic and adding tests. The `--template-dir` is a directory for Githooks to write its own template files to, it is **NOT** necessarily the same as `git config --global init.templateDir` . 

If you don't pass this in, Githooks might use `init.templateDir` if it's set and has a `hooks` directory. (see https://github.com/rycus86/githooks/blob/master/install.sh#L3724)

If you do pass it in, the installer will use that directory to copy the templates to (and from) but it won't mess with any of your existing Git config (see https://github.com/rycus86/githooks/blob/master/install.sh#L3692), e.g. with the next install you'll have to specify `--template-dir` again or let Githooks create a new template directory (which will be set to `init.templateDir`) - see https://github.com/rycus86/githooks/blob/master/install.sh#L3753 or https://github.com/rycus86/githooks/blob/master/install.sh#L3897

So in principle:
- if you pass in a template dir, we'll use that for reading stuff
[...]

PR #71, opened 2019-11-25
Refactoring: return values, template dir, warnings

gabyx:


install.sh
```diff
@@ -3778,24 +3794,22 @@ find_git_hook_templates() {
 ############################################################
 mark_directory_as_target() {
     TARGET="$1"
-    if [ "$TARGET" = "" ]; then
+    if [ -z "$TARGET" ]; then
         return
     fi
 
-    if [ "$2" != "" ]; then
```
gabyx: Adding $2 so early in the function is not correct here. See the function description. The logic to be correct  should in a fashion similar to my solution

uninstall.sh
```diff
@@ -396,6 +396,15 @@ uninstall() {
     # Uninstall all cli
     uninstall_cli
 
+    if using_hooks_path; then
```
gabyx: Moved that up, here in favor of:
1. clean dependant config stuff
2. clean all githooks config variables...

gabyx:
See the #70 for a description and example why this PR...

rycus86:
Hey! Sorry, haven't had a chance to review this yet. I'll try to get to it this week, time permitting. Thanks!

gabyx:
Please already have a look at the proper ordered commits with the changes which are improvements and fixes and docs.
- --template-dir has not been tested and has some bugs.

I will later fix the tests!

gabyx:
This would be good to go! :+1: Arch tests fail to the same reason as we know -> Bug...

install.sh
```diff
@@ -1161,15 +1179,27 @@ set_githooks_directory() {
         git config --global githooks.useCoreHooksPath yes
         git config --global githooks.pathForUseCoreHooksPath "$1"
         git config --global core.hooksPath "$1"
+
+        CURRENT_CORE_TEMPLATE_DIR=$(git config --global init.templateDir)
+        # shellcheck disable=SC2012
+        if [ "$(ls -1 "$CURRENT_CORE_TEMPLATE_DIR/hooks" 2>/dev/null | wc -l)" != "0" ]; then
```
rycus86, line 1185: This will almost always be true for regular git installations, since that comes with a bunch of sample files in there.
I believe git will use `core.hooksPath` first if it's set, so masking the hooks in the template dir is probably not that big of a deal, than on the opposite side.
I'd remove this error as I don't think it would help people much to be honest.
What do you think @gabyx ?

rycus86:
Also wanted to ask here, if this is a bugfix, could you add a test that fails on master but works on this branch? The refactoring looks OK, but I'm not sure what it is fixing.

gabyx:
Its not fixing anything sorry should be a refactoring. There were however some small subtle loop holes in the code.

Von meinem iPhone gesendet

install.sh
gabyx: Ahh you are right. That error will then be always visible when somebody is using core.hooksPath with a normal installation. Not good hm... I will check with the git manual. or test it if `git config --global init.templateDir` always resolvs to the standard installation default... Good point

gabyx: OK, as the dokumentation says and what I see here on the terminal. `core.initTemplateDir` [--local, --global, --system] is simply set by the user, it does not resort to the default location of a git install. So this error would be helpful I think.
We should also test for the "$GIT_TEMPLATE_DIR" as this could be set as well...

https://git-scm.com/docs/git-init

rycus86: Oh, did I confuse this with the default path at `/usr/share/git-core/templates` ?
> the init.templateDir configuration variable; or
> the default template directory: /usr/share/git-core/templates.

If this is the case on popular environments, then I'm fine with it.
- OS X (not sure how we can test this)
- Windows / Git Bash - can you test this?
- Debian / Ubuntu / Arch maybe

gabyx: Yes sorry variable naming is bad, changed it in my next commit.

gabyx: This warning is exactly what it says :-). I am not so eager to write a test just for this warning. ;-) Hm... or what did you thing should be tested?

rycus86: Whether `init.templateDir` is set on a default installation. If you're confident it's not then OK, for some reason I thought it is set on Debian or Ubuntu. Let me see if I can check.

gabyx: core.hooksPath is a complete override for anything :-) which gets installed with any kind of template dir used for git init 
https://git-scm.com/docs/git-init

rycus86: OK, I was wrong, it uses that default path on Ubuntu and Debian it seems. I guess it will be the same on Git Bash, no idea about OS X and homebrew, but let's assume it'll work it out without setting `init.templateDir`?

install.sh
```diff
@@ -1161,15 +1179,43 @@ set_githooks_directory() {
...
+
+        # shellcheck disable=SC2012
+        if [ -d "$GIT_TEMPLATE_DIR" ] &&
+            [ "$(ls -1 "$GIT_TEMPLATE_DIR/hooks" 2>/dev/null | wc -l)" != "0" ]; then
+            echo "! The environment variable \`GIT_TEMPLATE_DIR\` is currently set to" >&2
+            echo "  \`$GIT_TEMPLATE_DIR\`" >&2
+            HOOKS_GET_IGNORED=1
+        fi
+
+        if [ "$HOOKS_GET_IGNORED" = 1 ]; then
```
rycus86, line 1201: Nice one!
Some minor edit suggestion perhaps:
```
            echo "  and it contains Git hooks which get installed," >&2
            echo "  but will be ignored because \`core.hooksPath\` is also set." >&2
            echo "  It is recommended to either remove the files, or run the Githooks" >&2
            echo "  installation without the \`--use-core-hookspath\` parameter." >&2
            unset HOOKS_GET_IGNORED
```

We're not failing the install here, right? In which case the above changes should soften the message to a recommendation, instead of sounding like something you have to do.

install.sh
```diff
@@ -141,13 +141,6 @@ is_running_internal_install() {
 ############################################################
 # Set up variables based on command line arguments.
 #
-# Sets ${DRY_RUN} for --dry-run
```
rycus86, line 144: Why did these go away btw?

rycus86:
Had a comment on the message, plus you'll need to resolve the merge conflict, but otherwise looks ready to merge for me! 👍

install.sh
gabyx: On Linux (Ubuntu here) its available, I mean we only cover the specs of git, we also only try to use it if its available `mark_directory_as_target "/usr/share/git-core/templates/hooks" && return 0` 
and I dont know what how this review thread is related to the code you tagged?

install.sh
gabyx: Ups should be = "1"

install.sh
gabyx: Ohh, thx. Good question...

gabyx: Ahh because we would need to list all of them which is tediuous

gabyx: I though ahh comon drop it, its parsing args of course we set the variables and we can see it :-)=

install.sh
gabyx: Jeah I thought so too, will change that

install.sh
rycus86: OK, I'm not too fussed either way, so leave it if you changed it already.

install.sh
rycus86: I initially thought some Git installations come with a global `init.templateDir` set, but it turns out I was wrong, that's where this thread came from. It sounds like it's all good, nothing to change here. 👍

gabyx:
Rebased and "warning -> to recommendation" fix

rycus86:
Thanks for getting this through!

PR #72, opened 2019-12-12
Bugfix: env -i is too restrictive. 

gabyx:
Can we change that? 
`git hooks shared update` fails with `env: ‘git’: No such file or directory`
on windows. The reason for this change to `env -i` was because of defined `GIT_DIR` and stuff...
I changed it to a less harsh method. I also added to ignore running any hooks.

Thanks for merging. Hopefully all tests still work with that.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 1912.042217-43ad0f
+# Version: 1912.122150-8769e7
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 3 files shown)


rycus86:
Thanks for this! Don't we have a similar block in `base-template.sh`? Sounds like that might need a similar change?

rycus86:
Tests work by the way but Travis integration is a bit broken... https://travis-ci.org/rycus86/githooks/builds/624250703

I need to move off it eventually, just can't find the time.

gabyx:
ups yes indeed. thats why sharing general functionality wouldnt be the dumbest. copy paste is the dead of every repo

Von meinem iPhone gesendet

PR #76, opened 2020-01-30
Bugfix: Finding git directories.

gabyx:
To review. Not yet perfect.

```
GIT_DIR=$(cd "$EXISTING" && GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --git-dir)
```
needs revisit. Return relative path??

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 2001.241152-6f44a6
+# Version: 2001.301056-ee0376
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -28,7 +28,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2001.241152-6f44a6
+# Version: 2001.301056-ee0376
 
 #####################################################
 # Execute the current hook,
@@ -967,7 +967,7 @@ CLI_TOOL_CONTENT='#!/bin/sh
 # See the documentation in the project README for more information,
 #   or run the `git hooks help` command for available options.
 #
-# Version: 2001.241152-6f44a6
+# Version: 2001.301056-ee0376
 
 #####################################################
 # Prints the command line help for usage and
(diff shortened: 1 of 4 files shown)


install.sh
```diff
@@ -4101,6 +4101,47 @@ setup_automatic_update_checks() {
...
+    for EXISTING in $REPOSITORY_LIST; do
+        unset IFS
+
+        if [ -f "$EXISTING" ]; then
+            # Strip HEAD file
+            EXISTING=$(dirname "$EXISTING")
+        fi
+
+        # Go to the root git dir (works in bare and non-bare repositories)
+        GIT_DIR=$(cd "$EXISTING" && GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --git-dir 2>/dev/null)
```
rycus86: Is the `GIT_DISCOVERY_ACROSS_FILESYSTEM=0` necessary here?

install.sh
```diff
@@ -4101,6 +4101,47 @@ setup_automatic_update_checks() {
...
+        if [ -n "$GIT_DIR" ] && ! echo "$EXISTING_REPOSITORY_LIST" | grep -q "$GIT_DIR"; then
+            EXISTING_REPOSITORY_LIST=$(printf '%b' "$GIT_DIR\n$EXISTING_REPOSITORY_LIST")
```
rycus86: This should probably be `printf 'pattern' "$VAR1" "$VAR2"`

rycus86:
Do we have any other is_git_dir checks anywhere? Might be good to keep them consistent and doing the same thing.

rycus86:
Thanks a lot for looking into this btw!

install.sh
gabyx: I would put it in, because someone accidentally set that, I just disable it.

install.sh
gabyx: printf not so portable `printf "%b" "a\nb"`, I changed to `echo` with newline in the script

gabyx:
Please see the changes now.

Btw: Would it make sense to collaborate all copy&paste functions into a `general.sh` which we also inject during commiting. so we only need to fix bugs once, would you mind refactoring this? Would help in the future.

rycus86:
Changes look good, just waiting for Travis then I'll merge it.

I'll think about the shared snippet to insert, it's not only a good way to share code but also to share bloat that some scripts wouldn't have needed otherwise. 🙂

gabyx:
>Do we have any other is_git_dir checks anywhere? Might be good to keep them consistent and doing the same thing.

we only have `is_git_repo` tests and these are ok I think.
Other git_dir checks I dont know

gabyx:
I have not tested it by me :-) (on windows, unfortunately.)
If it does not work, I check tonight...

PR #77, opened 2020-01-31
Critical Bugfix: Correct wrong git-dir in install.

gabyx:
**Criticall fix. The test should work now.**
https://travis-ci.org/rycus86/githooks/builds/644352092
Except coverage which needs fixing later...

Made some stupid error. As you said, there was still a place where I forgot to check the correct git-dir.
Changed also stupid variable names `GIT_DIR` to `REPO_GIT_DIR`.

Sorry for the mistakes...

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 2001.301056-ee0376
+# Version: 2001.311420-ff1e90
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -28,7 +28,7 @@ BASE_TEMPLATE_CONTENT='#!/bin/sh
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2001.301056-ee0376
+# Version: 2001.311420-ff1e90
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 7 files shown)


install.sh
```diff
@@ -4343,11 +4343,10 @@ install_into_registered_repositories() {
             fi
 
             IFS="$IFS_NEWLINE"
-            while read -r INSTALLED_REPO; do
+            while read -r REPO_GIT_DIR; do
                 unset IFS
 
-                EXISTING_REPO_ROOT=$(dirname "$INSTALLED_REPO")
-                install_hooks_into_repo "$EXISTING_REPO_ROOT"
+                install_hooks_into_repo "$REPO_GIT_DIR"
```
gabyx: Thats the ugly bug!!!

tests/exec-tests.sh
```diff
@@ -19,6 +19,8 @@ ADD tests/exec-steps.sh tests/${STEPS_TO_RUN} /var/lib/tests/
 
 # Do not use the terminal in tests
 RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
+# We overwrite the download to use the current install.sh
```
gabyx: Really hand fix which should have been in a long time already...

rycus86:
The changes look good, but there are now 14 tests failing in the coverage build instead of just the last single one. How confident are you that this doesn't break those others? Failures seem related to LFS, shared hooks, etc. See https://travis-ci.org/rycus86/githooks/jobs/644352108

rycus86:
Merged for the bugfix, but please check the above if you can. Thanks!

gabyx:
Yeah, will have a look at it, strange

Von meinem iPhone gesendet

gabyx:
Ok, the problem (:sweat: ) is the to old git installation in `kcov:kcov`.
and still some error in my latest fix. I try to make another pull request. Maybe lets review that a bit more proper :)

PR #78, opened 2020-02-02
Bugfix/fix coverage

gabyx:
Some fixes.

- [x] old git installation in kcov -> fix git command to support all version
- [x] MOCK_DOWNLOAD in tests to be able to mock all downloads for locally testing (no artefacts over the master branch). `coverage.sh` uses always the MOCK_DOWNLOAD, because its not so relevant there. For `exec-tests.sh` ist opt-in!
- [x] check kcov output, all should pass now except 116 which needs fixes.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3515,7 +3515,7 @@ execute_installation() {
 
     if ! should_skip_install_into_existing_repositories; then
         if is_single_repo_install; then
-            CURRENT_GIT_DIR=$(git rev-parse --git-common-dir 2>/dev/null)
+            CURRENT_GIT_DIR=$(cd "$(git rev-parse --git-common-dir 2>/dev/null)" && pwd)
             install_hooks_into_repo "$CURRENT_GIT_DIR" || return 1
         else
             if ! is_update_only; then
@@ -4130,11 +4130,9 @@ find_existing_git_dirs() {
         # Try to go to the root git dir (works in bare and non-bare repositories)
         # to neglect false positives from the find above
         # e.g. spourious HEAD file or .git dir which does not mark a repository
-        REPO_GIT_DIR=$(cd "$EXISTING" && GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --absolute-git-dir 2>/dev/null)
-        # Convert the path to the convention this shell uses
-        # (e.g. on windows the above gives windows paths)
-        REPO_GIT_DIR=$(cd "$REPO_GIT_DIR" && pwd)
-        if [ -n "$REPO_GIT_DIR" ] && ! echo "$EXISTING_REPOSITORY_LIST" | grep -q "$REPO_GIT_DIR"; then
+        REPO_GIT_DIR=$(cd "$EXISTING" && cd "$(GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --git-dir 2>/dev/null)" && pwd)
+
+        if [ -d "$REPO_GIT_DIR" ] && ! echo "$EXISTING_REPOSITORY_LIST" | grep -q "$REPO_GIT_DIR"; then
             EXISTING_REPOSITORY_LIST="$REPO_GIT_DIR
 $EXISTING_REPOSITORY_LIST"
         fi
@@ -4399,6 +4397,7 @@ install_hooks_into_repo() {
         fi
 
         TARGET_HOOK="${TARGET}/hooks/${HOOK_NAME}"
+        echo "INstalled to $TARGET_HOOK"
 
         if [ -f "$TARGET_HOOK" ]; then
             grep 'https://github.com/rycus86/githooks' "${TARGET_HOOK}" >/dev/null 2>&1
@@ -4421,6 +4420,7 @@ install_hooks_into_repo() {
 
         if echo "$BASE_TEMPLATE_CONTENT" >"$TARGET_HOOK" && chmod +x "$TARGET_HOOK"; then
             INSTALLED="yes"
+            echo "INstalled to $TARGET_HOOK"
         else
             HAD_FAILURE=Y
             echo "! Failed to install $TARGET_HOOK" >&2
--- a/install.sh
+++ b/install.sh
@@ -112,9 +112,9 @@ set_main_variables() {
     HOOK_FOLDER=$(dirname "$0")
     ACCEPT_CHANGES=
 
-    CURRENT_GIT_DIR=$(git rev-parse --git-common-dir)
-    if [ "${CURRENT_GIT_DIR}" = "--git-common-dir" ]; then
-        CURRENT_GIT_DIR=".git" # reset to a sensible default
+    CURRENT_GIT_DIR=$(git rev-parse --git-common-dir 2>/dev/null)
+    if [ ! -d "${CURRENT_GIT_DIR}" ]; then
+        echo "! Hook not run inside a git repository" >&2 && exit 1
     fi
 
     load_install_dir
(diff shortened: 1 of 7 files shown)


gabyx:
I try to correctly mock the download but test 094 still downloads the master I think. pretty weird. will work on it later

gabyx:
I try to correctly mock the download but test 094 still downloads the master I think. pretty weird. will work on it later

rycus86:
Thanks a lot for looking into this!
Last time I tried to update kcov, I found that lots have changed in the next version, so I gave up on updating it at the time.

install.sh
```diff
@@ -4130,11 +4130,9 @@ find_existing_git_dirs() {
         # Try to go to the root git dir (works in bare and non-bare repositories)
         # to neglect false positives from the find above
         # e.g. spourious HEAD file or .git dir which does not mark a repository
-        REPO_GIT_DIR=$(cd "$EXISTING" && GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --absolute-git-dir 2>/dev/null)
-        # Convert the path to the convention this shell uses
-        # (e.g. on windows the above gives windows paths)
-        REPO_GIT_DIR=$(cd "$REPO_GIT_DIR" && pwd)
-        if [ -n "$REPO_GIT_DIR" ] && ! echo "$EXISTING_REPOSITORY_LIST" | grep -q "$REPO_GIT_DIR"; then
+        REPO_GIT_DIR=$(cd "$EXISTING" && cd "$(GIT_DISCOVERY_ACROSS_FILESYSTEM=0 git rev-parse --git-dir 2>/dev/null)" && pwd)
```
rycus86: Is this something we should get into master regardless of the coverage fix?
Do you know from which version does git have the new parameter that kcov didn't like?

rycus86: Also, should this be changed in the uninstall script too?

gabyx:
> Last time I tried to update kcov, I found that lots have changed in the next version, so I gave up on updating it at the time.
Tried it too, but lets do that in another PR maybe...

tests/exec-tests.sh
```diff
@@ -19,14 +26,20 @@ ADD tests/exec-steps.sh tests/${STEPS_TO_RUN} /var/lib/tests/
...
     sed -i 's|ACCEPT_CHANGES=|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's%read -r "\$VARIABLE"%eval "\$VARIABLE=\\\\\$\$(eval echo "\\\\\$VARIABLE")" # disabled for tests: read -r "\$VARIABLE"%' /var/lib/githooks/base-template.sh
-    
+
+RUN if [ -n "$MOCK_DOWNLOAD" ]; then \\
+    # We overwrite the download to use the current install.sh in all scripts
+    sed -i -E 's@(curl|wget).*(DOWNLOAD_URL|OUTPUT_FILE).*(DOWNLOAD_URL|OUTPUT_FILE).*@cp -f /var/lib/githooks/install.sh "\$OUTPUT_FILE"@g' \\
+        /var/lib/githooks/install.sh \\
+        /var/lib/githooks/cli.sh  \\
+        /var/lib/githooks/base-template.sh \\
```
rycus86: Looks like this line needs a semicolon

gabyx: True, fixed!

gabyx:
Now, all coverage works except 116 and I still dont know why, strange... kcov is doing strange things...
even the output is scrambled, :-|. Really cumbersome...

install.sh
gabyx: Dont know exactly, but its still good to not really use the newest stuff if we can avoid it.

gabyx:
Is there any consense about using `git rev-parse --commo-git-dir` versus `--git-dir`.
Whould we use everywhere `--commo-git-dir`

rycus86:
> Now, all coverage works except 116 and I still dont know why, strange... kcov is doing strange things...
> even the output is scrambled, :-|. Really cumbersome...

Yeah, that output gets scrambled unfortunately, never really worked out how to fix it properly.
If you look at the generated coverage report in HTML, you can see which lines were executed vs which ones did you expect to be executed.

> Is there any consense about using `git rev-parse --commo-git-dir` versus `--git-dir`.
> Whould we use everywhere `--commo-git-dir`
> Git has to many behaviors :-), not a simple tool :-P

Nope, I have no personal preference, except whichever existed first, so we don't limit compatibility with older Git versions if not necessary.

gabyx:
Ok. I think this PR is good to go. 116 is still a mystery. Kcov is strange because sometimes not the whole output is shown and it stops suddenly (or just because we cannot see the whole output) and its pretty unobvious where it stopped. I will see if I can detect that in the report. A `set -x` lets everything go nuts :-)... Updating is an option? If I have time I will look into this.

gabyx:
https://travis-ci.org/rycus86/githooks/builds/645188064

gabyx:
Maybe we can leave test 116 out of the coverage, just to make it build and let the users know the build is fine. 116 should be fixed, will look into it soon hopefully with more details.

install.sh
rycus86: Why do we need `GIT_DISCOVERY_ACROSS_FILESYSTEM=0` here, or alternatively why don't we need it elsewhere?
I'd just drop it to be honest.

install.sh
```diff
@@ -4399,6 +4397,7 @@ install_hooks_into_repo() {
         fi
 
         TARGET_HOOK="${TARGET}/hooks/${HOOK_NAME}"
+        echo "INstalled to $TARGET_HOOK"
```
rycus86: Debugging leftover? Can you please remove this?

install.sh
```diff
@@ -4421,6 +4420,7 @@ install_hooks_into_repo() {
 
         if echo "$BASE_TEMPLATE_CONTENT" >"$TARGET_HOOK" && chmod +x "$TARGET_HOOK"; then
             INSTALLED="yes"
+            echo "INstalled to $TARGET_HOOK"
```
rycus86: Can you remove this too please?

uninstall.sh
```diff
@@ -307,8 +292,10 @@ unregister_repo_for_autoupdate() {
 #   0 on success, 1 on failure
 ############################################################
 uninstall_from_current_repository() {
-    if ! is_running_in_git_repo_root; then
-        echo "The current directory ($(pwd)) does not seem to be the root of a Git repository!" >&2
+    CURRENT_GIT_DIR=$(cd "$(git rev-parse --git-common-dir 2>/dev/null)" && pwd)
+
+    if ! is_git_repo "$(pwd)" || [ ! -d "$CURRENT_GIT_DIR" ]; then
```
rycus86: Wouldn't `! is_git_repo "$CURRENT_GIT_DIR"` be enough here?

rycus86: Oh the git dir will point to a `.git` directory normally?
In that case, it feels like the `git rev-parse` check should be inside the `is_git_repo` function.

rycus86:
Added some comments, but looks good in general. We can get this in and leave test 116 as is for now, just look at the comments please. Thanks!

install.sh
gabyx: Thx

uninstall.sh
gabyx: Correct. if we are inside a git repo, he directory should be valid! :-)

gabyx: Note: the `if ! is_git_repo "$(pwd)" ` is only to check if `uninstall_from_current_repository` is really run from inside a git repository, then also `CURRENT_GIT_DIR` is set and valid. I will change that

gabyx:
I did some review cleanup, and changed every relevant point to `--git-common-dir` as its the right thing to do. And I have seen you already did some tests for git worktrees :+1: 

Could you please review the last changeset. If you are happy, maybe you can directly merge it, without having me to squash it :-)

gabyx:
I will also try to replace all `cd "path" && git ...` to `git -C "path"` quickly.

gabyx:
All tests work. Covarage still the same.

uninstall.sh
gabyx: See latest review commits.

rycus86:
Thanks a lot!

Issue #84, opened 2020-04-16
Please never implicitly accept a trust prompt


matthijskooijman:
The README says:

> There is a caveat worth mentioning: if a terminal (tty) can't be allocated, then the default action is to accept the changes or new hooks. Let me know in an issue if you strongly disagree, and you think this is a big enough risk worth having slightly worse UX instead.

I strongly disagree :-)

Because hooks can run arbitrary code, this is a significant security issue and should never be done without explicit consent. If a value cannot be read from the user, the request should be rejected IMHO. The script could maybe print a command you could run to explicitly allow the hooks in the current dir (which seems to be just `githooks accept`) and maybe sleep for a bit to allow aborting if there *is* a user looking, but then continue without the hook.

Currently, the prompt also shows Y as the default (i.e. what would be used if the user pressed "enter" without entering anything). I would also suggest (but less strongly) to change that and require *explicit* consent from the user. Pressing enter just a bit too soon is easy to do, and can have security implications here. It would be good to be explicit about this: When a user presses enter directly, show the prompt again and ask for an explicit answer. This needs distinguishing an "empty response" from "no response", but I think the `read` exit status should allow this (might need some fiddling in `show_prompt()` to actually return the right value, but maybe it already works because `read` is the last command, see also https://github.com/rycus86/githooks/pull/81#issuecomment-614782632).

At the very least this should be configurable, and I also strongly suggest making the default the secure option. If this is annoying for users and they trust their repository, they can always use the "trust" option to get rid of the annoying prompts.

gabyx:
OK, this seems feasible. So the following code in `base_template.sh` is for reconsideration:
```shell
# Try to read from `dev/tty` if available.
    # otherwise just read from stdin.
    printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
    if [ -t 0 ]; then
        # shellcheck disable=SC2229
        read -r "$VARIABLE" </dev/tty
    else
        # shellcheck disable=SC2229
        read -r "$VARIABLE"
    fi
```
and the two lines 
```shell
MESSAGE="$(printf "%s\n%s" "$MESSAGE: $HOOK_PATH" "  Do you accept the changes?")"
show_prompt ACCEPT_CHANGES "? $MESSAGE" "(Yes, all, no, disable)" "Y/a/n/d" "Yes" "All" "No" "Disable"
```

gabyx:
I agree. :-) The thing is if we strong need a user input from
[...]

PR #85, opened 2020-04-16
Bugfix: No prompt if no controlling terminal could be detected :tada:

gabyx:
By https://github.com/rycus86/githooks/pull/81#issuecomment-614782632
this fix should fix the logic when we read from /dev/tty for getting an answer to a prompt.
We dont modify the variable read into if we could not read something from a controlling terminal:

@matthijskooijman Shouldn't we only (if and only if) read from /dev/tty when stdin is also a tty meaning the change should be maybe even more strict: 
```shell
if [-t 0] && [ -t 1 ]; then
        # shellcheck disable=SC2229
        read -r "$VARIABLE" </dev/tty
    fi
```

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2004.162028-00b20e
+# Version: 2004.162031-c978c3
 
 #####################################################
 # Execute the current hook,
@@ -764,8 +764,8 @@ show_prompt() {
     # However, only do this when stdout *is* a tty, otherwise it is
     # likely we have no controlling terminal and reading from /dev/tty
     # would fail with an error.
+    printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
     if [ -t 0 ] && [ -t 1 ]; then
-        printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
         # shellcheck disable=SC2229
         read -r "$VARIABLE" </dev/tty
     fi
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2004.162018-62f173
+# Version: 2004.162028-00b20e
 
 #####################################################
 # Execute the current hook,
@@ -764,7 +764,7 @@ show_prompt() {
     # However, only do this when stdout *is* a tty, otherwise it is
     # likely we have no controlling terminal and reading from /dev/tty
     # would fail with an error.
-    if [ -t 1 ]; then
+    if [ -t 0 ] && [ -t 1 ]; then
         printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
         # shellcheck disable=SC2229
         read -r "$VARIABLE" </dev/tty
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2004.162031-c978c3
+# Version: 2004.170929-8e4d56
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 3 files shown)


gabyx:
No controlling terminal is available for example when we commit by a GUI tool, I think vscode should work...?

gabyx:
Lets see if all tests work, by me everythin passes for `test-alpine-curl.sh`

matthijskooijman:
> Shouldn't we only (if and only if) read from /dev/tty when stdin is also a tty meaning the change should be maybe even more strict:

No, as I mentioned in #81, stdin *is* never a tty when run from git, and *if it is*, the /dev/tty redirect is not even needed.

matthijskooijman:
Testing *without* controlling tty will probably happen already in the testsuite (since I suspect all of that runs without tty). It would be interesting to test this *with* a tty as well, which would I think require creating a PTY pair so you can run git on one half and send some "interactive" input to the other half.

I googled around a bit on how to do this from bash, which seems non-trivial (a C program needs to do all kinds of setup for this). However, it seems that you can let socat do the heavy lifting, according to https://stackoverflow.com/a/19848887/740048

gabyx:
Ok to summarize to make sure I understand correctly:
My input output know how is somewhat newbie:
when running any hook we have 
- `stdin` (dont read from that for user input, because of git) 
- `stdout` 
- and maybe a controlling terminal which we can get by `/dev/tty`.  

As you experienced it seems that `git` attaches `/dev/null` or a `pipe` mostly to `stdin`. 
- Reading from `stdin` is never good in that way (because of info which is sent by a pipe for example, I hope this `stdin` gets inherited to the hooks githooks runs, this should probably be the case...)
The test `[ -t 0 ]` in https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html
tests if a file descriptor is associated with a terminal (thats the same as a "controlling terminal" hopefuly). This does not help also.

So, when I want to show a user prompt and answer from the user I can only do this by `/dev/tty`.  There is a another POSIX command `tty` (https://pubs.opengroup.org/onlinepubs/9699919799/utilities/tty.html#tag_20_135) 
which returns 0 if `stdin` is a terminal, but what does that help.

SO the only way is (as you mentioned) to open the file /dev/tty check if is valid.

gabyx:
I reverted to the initial command by @rycus86 which is more what we want.
`read -r "$VARIABLE" </dev/tty 2>/dev/null`
and fixes the error message saying `no file descriptor` etc...

matthijskooijman:
Your summary seems correct. Your proposal of just always reading from `/dev/tty` and masking the error would also seem useful to me (though there is a chance of masking *other* errors that might have been useful). You did not update the comment yet, so if this is merged, that should be done first.

I'm curious: When `/dev/tty` is not available, what does the `read` command return? i.e. what do you get when you put `echo $?` directly after the read? Knowing that might be helpful in relation to #84.

base-template.sh
```diff
@@ -759,15 +759,18 @@ show_prompt() {
...
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    if [ -t 0 ]; then
-        # shellcheck disable=SC2229
-        read -r "$VARIABLE" </dev/tty
-    else
-        # shellcheck disable=SC2229
-        read -r "$VARIABLE"
-    fi
+    # shellcheck disable=SC2229
+    read -r "$VARIABLE" </dev/tty 2>/dev/null
```
rycus86, line 769: Why do we need to suppress the error here? Could we just let it through? It might help working out problems better.

rycus86: @gabyx can you have a look at this please? This PR should be good to go otherwise.

matthijskooijman: If you don't suppress this error, I think you're essentially just reverting #81 and you are back to the problem described there (which I *think* is only that an error message is shown, I do not think that #81 actually fixed anything otherwise).

rycus86: Ah yes, I see. Not super awesome to suppress the error but I guess let's try and see. @gabyx happy for this PR to go in? We're not waiting on anything else for this, right?

gabyx:
@rycus86 yes this PR can go in :-). as you mentioned: suppressing the error is #81 eseentially. lets leave it in, should be fine I think.

base-template.sh
matthijskooijman: (Hm, started typing this comment before the merge and then got called away. In case it's still usedful, I'll finish it anyway)

The alternative that also seems reasonable to me (which I suggested before), is to check whether stdout is a tty (`[ -t 1 ]`) and if not, just not read at all (maybe not even show the prompt) and return an error. This probably ties in with #84 a bit as well.

Or, I realize now, you could also maybe first try opening /dev/tty separately and supress output separately there. E.g. `if true < /dev/tty 2> /dev/null; then` might work to check if opening `/dev/tty` would work without suppressing (much) other errors. Ideally you would just do a libc `open` call and check the return code (for a specific error code, `ENODEV` I  think), but I don't think sh/bash allows that.

rycus86: > Or, I realize now, you could also maybe first try opening /dev/tty separately and supress output separately there. E.g. if true < /dev/tty 2> /dev/null; then might work to check if opening /dev/tty would work without suppressing (much) other errors.

Hm, that sounds good, it might be worth to try. @gabyx any chance you could test it in the environment this was originally broken for you?

Issue #83, opened 2020-04-16
Simplify installation?


matthijskooijman:
I'm looking at using githooks (not using it yet) and reviewing the install script, but it looks like the installation is a bit complicated, more than I would think is needed. I suspect that most of this might be so that `install.sh` becomes a self-contained file that can be run through a `sh $(curl ...)` command? However, I don't quite like that way of installing and updating stuff, I'd rather just clone this repository and update things by pulling. Is this, or can this be, a supported usecase?

I originally wanted to remark that the contents of `base-template.sh` are duplicated inside `install.sh`. Before I realized install.sh should be self-contained, it seemed sane to just let `install.sh` read `base-template.sh` at runtime, but that won't work here, I guess. I do see that there is a pre-commit hook that actually *updates* `install.sh` with the content from some other files.

Anyway, ignoring the self-contained scenario for a short while, here's how my ideal flow would look like:
 - Clone this repo somewhere in my homedir, let's say `~/githooks`
 - Configure git (manually is fine) to point the hookdir (i.e. `core.hooksPath`) towards this checkout (probably a subdir, e.g. `core.hooksPath = ~/githooks/hooks`). Alternatively, you can point the hook template config value to this directory.
 - Inside this subdir, there would be the `base-template.sh` file (maybe with better name) to do the actual work, and have small files for each supported hook, that do nothing more than calling `base-template.sh` (taking care not to change `$0`, or otherwise passing some variables like `HOOK_NAME` explicitly). This could be done with symlinks, but that is not portable to Windows I think.
 - Then, symlink `cli.sh` into my `$PATH` so I can call that.

Putting it like this, this might mean this does not actually need the install script at all, so this approach could likely co-exist nicely with the existing auto-generated self-contained install script maybe?

matthijskooijman:
I just tried this and it actually seems to work quite well so far. I:
 - Created a git clone in ~/.local/etc/
 - Created a `~/.githook/hooks` subdirectory
 - Inside that subdir, I ran:
   ```
   HOOKS="applypatch-msg commit-msg post-applypatch post-checkout post-commit post-merge post-receive post-rewrite post-update pre-applypatch pre-auto-gc pre-commit prepare-commit-msg pre-push pre-rebase pre-receive push-to-checkout sendemail-validate update"
   for i in $HOOKS; do ln -s ../base-template.sh $i;done
   ```
   This creates the hooks to point git at, which all just run the base-template through a symlink.
[...]

PR #86, opened 2020-04-18
Feature: Clone a release of githooks into the install folder

gabyx:
Ok, this feature simplifies the githooks implementation a lot, by only using git, no curl/wget.

When installing with `install.sh [--release-clone-url="..."]` we clone a githooks version into `$INSTALL_DIR/release` which we use for further updates. These changes are backward compatible, (all tests work) and has the following major benefits:

- no more curl/wget, only git
- less tests
- no self-contained install - not neccessary anmore
- less commit fiddling (development hooks)
- **download tool** removed - not necessary anymore
- all curl/wget tests removed - not necessary anymore
- `$INSTALL_DIR/bin/cli.sh` dispatches to `$INSTALL_DIR/release/cli.sh`

Everything works as is.


gabyx:
@rycus86 Could you please give this (IMO beneficial improvements (its now like a `homebrew update` install :-P) a review.

matthijskooijman:
Thanks for picking this up! I am not familiar enough with the code (and don't have enough time now) to do a thorough review, but I do have some comments.

First, responding to https://github.com/rycus86/githooks/issues/83#issuecomment-615943130, which I think applies to this implementation:

> so install.sh would only basically clone this repo from a branch releases or s)

Oh, that's also an interesting thought, might even simplify things even more (but might require a working git client, or at least a working unzip if you let install.sh download the zip from github).


> base_template.sh gets copied for each hook into the git repo or it does not if we have set a core.hooksPath and we installed to use the core.hooksPath mode. Here might be room for improvement, by not copying the whole file, but this is another story...

I think using symlinks, or a one-liner script that execs the original base-template.sh would be a good improvement here. That would also mean that updates are automatic, solving:

>     We bookkeep a "cat ~/.githooks/autoupdate/registered" list with all repos (which are not single installs) which will be updated.

(except when a new hook is added, or the clone is moved, then you *would* need to update all repos)

Then, responding to this PR:

> no more curl/wget, only git

If you really intend to not use curl/wget but only git, that means the user must start with a git clone. So then why would the install script then do *another* clone (AFAIU it does)? I can imagine it does not and you always install using git, *or* you (also) support a curl/wget install where `install.sh` *does* do a git clone?

Looking at the changes, I see no addition of symlinks/single-line shims for each available hook, so I assume this means `install.sh` creates them? I would suggest adding them to the repo directly, so you can do a manual install by pointing the template or global hooks dir directly into this repo, without having to call install.sh?

rycus86:
To add to this, the whole project started as a script to set up your git template dir, so when creating new clones, they would automatically have the smarts to run hooks that you check into your repo. Plus it had to find existing repos on installing githooks, so if you already checked out a repo, hooks inside could start working. This is in line with how Git template directories work, and it's the reason why not everything is symlinked to a single set of files.

Another option that was added later is to support git core.hooksPath which does keep one set of files, so that's an option.

Getting the initial template  and install script files from this repo via git clone is pretty much the same as getting it with curl/wget, so I'm not sure why this would need to change now. Having said that, I kind of like the updates via git pull, so perhaps there's that bit to improve.

gabyx:
@matthijskooijman Your comments are for sure valuable, but as rycus said, the project has much more subtleties than from the first sight. As fas as "to be in controll" concerns: I think its not that opaque. The whole stuff is managed inside the install folder e.g. `~/.githooks`. The install.sh script also servers another purpose of only just cloneing a release, it manages other repos and also other already in place hooks as @rycus86 already mentioned and in the future (hopefully not) its possible to have backward compatibility stuff there too... so if we have an uninstall script which cleans git config variables`githooks.xxx` we definitely have a install script... and also this install script is the ground truth for updates too (and a handy way for inexperienced users just to set up everything under the install folder), in this PR we pull the release and execute it. @matthijskooijman: The procedure is more that you get the install.sh script (zip file, git clone if you like..., curl/wget) and then everything gets setup (git clone again, which might be counter intuitive...). However you have `git` installed and otherwise there is no reason to install this project at all! ;)

For this PR, I keep everthing which is not really important right now out of the discussion, which we rather leave for further improvement (which might really benefit, I mean @matthijskooijman tought about only having only a git repo (no install.sh)  would be nice, but I am not sure if that gonna work out....

gabyx:
@rycus86 : I will try to fix the travis stuff tonight. More or less the PR should work =)

gabyx:
All tests pass now. Review should be good to go :-) Thx @rycus86 .

tests/step-097.sh
```diff
@@ -19,7 +19,7 @@ fi
 mkdir -p /tmp/test097/.git/hooks &&
     cd /tmp/test097 &&
     git init &&
-    sh /var/lib/githooks/install.sh "$SINGLE" &&
+    sh /var/lib/githooks/install.sh $SINGLE &&
```
gabyx, line 22: Any wrong argument is rejected also empty ones -> safer...

rycus86:
> All tests pass now. Review should be good to go :-) Thx @rycus86 .

Thanks @gabyx ! I'll need to check out the code to properly review this change, but from a glance on my mobile, it looks all right. 🎉 I'll try to get it reviewed this weekend.

A question on updating to this from the "old" way, how's that work for people? Will it just switch over seamlessly?

gabyx:
I think, it should switch seamlessly: The scripts (cli.sh, hooks) get the install.sh from the github, then it will create the clone if its not there, I think in any cirumstance. I will check this path...

gabyx:
> I think, it should switch seamlessly: The scripts (cli.sh, hooks) get the install.sh from the github, then it will create the clone if its not there, I think in any cirumstance. I will check this path...

Checked: Should work :crossed_fingers:

.githooks/pre-commit/no-single-quote
```diff
@@ -1,10 +0,0 @@
-#!/bin/sh
```
rycus86, line 1: Can we keep this hook? I kind of like not having to think about quotes, and just disallow single quote in those files.
After this PR in, perhaps I'll check if the `install.sh` can also use double-quotes only.

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
-            curl -fsSL "$DOWNLOAD_URL" -o "$OUTPUT_FILE" >/dev/null 2>&1
-        elif wget --version >/dev/null 2>&1; then
-            wget -O "$OUTPUT_FILE" "$DOWNLOAD_URL" >/dev/null 2>&1
-        else
-            echo "! Cannot download file \`$DOWNLOAD_URL\` - needs either curl or wget" >&2
+############################################################
+# Clone the URL `$GITHOOKS_CLONE_URL` into the install
+# folder `$INSTALL_DIR/release` for further updates.
+#
+# Returns: 0 if succesful, 1 otherwise
```
rycus86, line 816:
```suggestion
# Returns: 0 if successful, 1 otherwise
```
Typo.

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "$INSTALL_DIR/release" >/dev/null 2>&1; then
+            echo "! Failed to remove an existing githook release repository" >&2
             return 1
         fi
     fi
 
-    # shellcheck disable=SC2181
+    echo "Installing a clone of \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```
rycus86, line 838:
```suggestion
    echo "Cloning \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+        --depth 1 \
+        --single-branch \
+        --branch "$GITHOOKS_CLONE_BRANCH" \
+        "$GITHOOKS_CLONE_URL" "$INSTALL_DIR/release" 2>&1)
+
+    #shellcheck disable=2181
     if [ $? -ne 0 ]; then
-        echo "! Cannot download file \`$DOWNLOAD_FILE\` - command failed" >&2
+        echo "! Cloning of repository \`$GITHOOKS_CLONE_URL\`"
+        echo "! to install directory \`$INSTALL_DIR/release\` failed with output: " >&2
```
rycus86, line 850:
```suggestion
        echo "! Cloning \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` failed with output: " >&2
```

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+        PULL_OUTPUT=$(
+            git -C "$CLONE_DIR" \
+                --work-tree="$CLONE_DIR" \
+                --git-dir="$CLONE_DIR/.git" \
+                -c core.hooksPath=/dev/null pull --all 2>&1
+        )
+
+        #shellcheck disable=2181
+        if [ $? -ne 0 ]; then
+            echo "! Pulling release clone in  \`$CLONE_DIR\` failed with:" >&2
```
rycus86, line 915:
```suggestion
            echo "! Pulling updates in \`$CLONE_DIR\` failed with:" >&2
```

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+    else
+        PULL_ONLY="false"
     fi
+
+    if [ "$PULL_ONLY" = "true" ]; then
+        PULL_OUTPUT=$(
+            git -C "$CLONE_DIR" \
+                --work-tree="$CLONE_DIR" \
+                --git-dir="$CLONE_DIR/.git" \
+                -c core.hooksPath=/dev/null pull --all 2>&1
```
rycus86, line 910: If we do a single branch clone with depth=1, shouldn't we also just `git pull origin <release-branch>` here?

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+            PULL_ONLY="false"
+        fi
+    else
+        PULL_ONLY="false"
     fi
+
+    if [ "$PULL_ONLY" = "true" ]; then
+        PULL_OUTPUT=$(
+            git -C "$CLONE_DIR" \
+                --work-tree="$CLONE_DIR" \
```
rycus86, line 908: Remind me again, why do we need so many flags on this pull?
Is it to stop the git variables from the current hook invocations leak into this other git invocation?

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+                --git-dir="$CLONE_DIR/.git" \
+                -c core.hooksPath=/dev/null pull --all 2>&1
+        )
+
+        #shellcheck disable=2181
```
rycus86, line 913:
```suggestion
        # shellcheck disable=SC2181
```

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+    echo "Installing a clone of \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
+
+    CLONE_OUTPUT=$(git clone \
+        -c core.hooksPath=/dev/null \
+        --depth 1 \
+        --single-branch \
+        --branch "$GITHOOKS_CLONE_BRANCH" \
+        "$GITHOOKS_CLONE_URL" "$INSTALL_DIR/release" 2>&1)
+
+    #shellcheck disable=2181
```
rycus86, line 847:
```suggestion
    # shellcheck disable=SC2181
```

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+    GITHOOKS_CLONE_URL=$(git config --global githooks.autoupdate.releaseCloneUrl)
+    GITHOOKS_CLONE_BRANCH=$(git config --global githooks.autoupdate.releaseCloneBranch)
+
+    if is_git_repo "$CLONE_DIR"; then
+        URL=$(git -C "$CLONE_DIR" config remote.origin.url)
+        BRANCH=$(git -C "$CLONE_DIR" symbolic-ref -q --short HEAD)
+
+        if [ "$URL" != "$GITHOOKS_CLONE_URL" ] ||
+            [ "$BRANCH" != "$GITHOOKS_CLONE_BRANCH" ] ||
+            git -C "$CLONE_DIR" diff --quiet; then
```
rycus86, line 898: This means we're going to `rm -rf` and clone over any local changes, right?
I wonder if we could try a clean merge and fail the installation if that doesn't work instead?
We could say, you're local install clone is dirty, either update manually or revert your local changes.
What do you think? It might be too much of a hassle, just wanted to throw it out there as a suggestion.

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+    CLONE_DIR="$INSTALL_DIR/release"
+    PULL_ONLY="true"
+
+    # We do a fresh clone if:
+    # - we have an existing clone AND
+    #   - url or branch of the existing clone does not match the settings OR
+    #   - the existing clone has modifications (should not be the case)
+    # - we dont have an existing clone
+
+    GITHOOKS_CLONE_URL=$(git config --global githooks.autoupdate.releaseCloneUrl)
```
rycus86, line 889: As a general note for the whole PR, could you please add quotes around all these assignments please?
`SOME_VAR="$(some command)"`

base-template.sh
```diff
@@ -774,61 +787,140 @@ show_prompt() {
...
+    GITHOOKS_CLONE_BRANCH=$(git config --global githooks.autoupdate.releaseCloneBranch)
+
+    if is_git_repo "$CLONE_DIR"; then
+        URL=$(git -C "$CLONE_DIR" config remote.origin.url)
+        BRANCH=$(git -C "$CLONE_DIR" symbolic-ref -q --short HEAD)
```
rycus86, line 894: Nice, didn't know about this one, always used `git rev-parse --abbrev-ref HEAD` instead (but according to SO, that has some issues sometimes)

cli.sh
```diff
@@ -1493,60 +1493,136 @@ call_script() {
...
-        else
-            echo "! Cannot download file \`$DOWNLOAD_URL\` - needs either curl or wget"
-            return 1
-        fi
+    if ! is_git_repo "$INSTALL_DIR/release"; then
+        clone_release_repository || return 1
     fi
 
-    # shellcheck disable=SC2181
+    #shellcheck disable=2154
```
rycus86, line 1547:
```suggestion
    # shellcheck disable=SC2154
```

Why is this necessary btw? We already use `$INSTALL_DIR` above.

cli.sh
```diff
@@ -1493,60 +1493,136 @@ call_script() {
...
+    GITHOOKS_CLONE_DIR="$INSTALL_DIR/release"
+
+    PULL_OUTPUT=$(git -C "$GITHOOKS_CLONE_DIR" --work-tree="$GITHOOKS_CLONE_DIR" --git-dir="$GITHOOKS_CLONE_DIR/.git" -c core.hooksPath=/dev/null pull 2>&1)
+    #shellcheck disable=2181
```
rycus86, line 1551:
```suggestion
    # shellcheck disable=SC2181
```

cli.sh
```diff
@@ -1493,60 +1493,136 @@ call_script() {
...
+    fi
+
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "$INSTALL_DIR/release" >/dev/null 2>&1; then
+            echo "! Failed to remove an existing githook release repository" >&2
+            return 1
+        fi
+    fi
+
+    echo "Installing a clone of \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```
rycus86, line 1605:
```suggestion
    echo "Cloning \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```

cli.sh
```diff
@@ -1493,60 +1493,136 @@ call_script() {
...
+        -c core.hooksPath=/dev/null \
+        --depth 1 \
+        --single-branch \
+        --branch "$GITHOOKS_CLONE_BRANCH" \
+        "$GITHOOKS_CLONE_URL" "$INSTALL_DIR/release" 2>&1)
+
+    #shellcheck disable=2181
+    if [ $? -ne 0 ]; then
+        echo "! Cloning of repository \`$GITHOOKS_CLONE_URL\`"
+        echo "! to install directory \`$INSTALL_DIR/release\` failed with output: " >&2
```
rycus86, line 1617:
```suggestion
        echo "! Cloning \`$GITHOOKS_CLONE_URL\`" to \`$INSTALL_DIR/release\` failed with output: " >&2
```

install.sh
```diff
@@ -3495,6 +34,14 @@ execute_installation() {
 
     load_install_dir
 
+    # Clone the repository to the install folder
+    # and run the install.sh from there.
+    if ! is_running_internal_install && ! is_update_only; then
```
rycus86, line 39: What's this block for, when would this `install.sh` not run from the clone repo?
On the first install from the old/current way?

install.sh
```diff
@@ -4046,8 +616,9 @@ setup_hook_templates() {
 ############################################################
 install_command_line_tool() {
     mkdir -p "$INSTALL_DIR/bin" &&
-        echo "$CLI_TOOL_CONTENT" >"$INSTALL_DIR/bin/githooks" &&
-        chmod +x "$INSTALL_DIR/bin/githooks" &&
+        echo "#!/bin/sh" >"$INSTALL_DIR/bin/githooks" &&
+        echo "sh '$INSTALL_DIR/release/cli.sh' \"\$@\"" >>"$INSTALL_DIR/bin/githooks"
+    chmod +x "$INSTALL_DIR/bin/githooks" &&
         git config --global alias.hooks "!$INSTALL_DIR/bin/githooks" &&
```
rycus86, line 622: Why not just set the `alias.hooks` to `$INSTALL_DIR/release/cli.sh'` instead?

install.sh
```diff
@@ -4604,6 +1175,127 @@ set_githooks_directory() {
...
+
+    if [ "$PULL_ONLY" = "true" ]; then
+        PULL_OUTPUT=$(
+            git -C "$CLONE_DIR" \
+                --work-tree="$CLONE_DIR" \
+                --git-dir="$CLONE_DIR/.git" \
+                -c core.hooksPath=/dev/null pull --all 2>&1
+        )
+
+        #shellcheck disable=2181
```
rycus86, line 1224:
```suggestion
        # shellcheck disable=SC2181
```

install.sh
```diff
@@ -4604,6 +1175,127 @@ set_githooks_directory() {
...
+        if [ $? -ne 0 ]; then
+            echo "! Pulling release clone in  \`$CLONE_DIR\` failed with:" >&2
```
rycus86, line 1226:
```suggestion
            echo "! Pulling release clone in \`$CLONE_DIR\` failed with:" >&2
```

install.sh
```diff
@@ -4604,6 +1175,127 @@ set_githooks_directory() {
...
+    fi
+
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "$INSTALL_DIR/release" >/dev/null 2>&1; then
+            echo "! Failed to remove an existing githook release repository" >&2
+            return 1
+        fi
+    fi
+
+    echo "Installing a clone of \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```
rycus86, line 1260:
```suggestion
    echo "Cloning \`$GITHOOKS_CLONE_URL\` to \`$INSTALL_DIR/release\` ..."
```

tests/coverage.sh
```diff
@@ -16,7 +16,8 @@ RUN echo 'deb http://deb.debian.org/debian stretch main' >> /etc/apt/sources.lis
     && (apt-get update || true) \
     && apt-get install -y git
 
-ADD base-template.sh install.sh uninstall.sh cli.sh .githooks/README.md /var/lib/githooks/
```
rycus86, line 19: Why did this have to change?

.githooks/pre-commit/no-single-quote
gabyx: Sure. !

tests/test-rules.sh
```diff
@@ -20,13 +20,10 @@ run_pre_commit_test() {
     fi
 }
 
-run_pre_commit_test no-single-quote
```
rycus86, line 23: Please keep it, see above.

uninstall.sh
```diff
@@ -412,6 +412,21 @@ uninstall_cli() {
...
+############################################################
+# Uninstall the release repository.
+#
+# Returns:
+#   None
+############################################################
+uninstall_release_repo() {
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "${INSTALL_DIR:?}/release" >/dev/null 2>&1; then
+            echo "! Failed to delete the githook release repository" >&2
```
rycus86, line 424:
```suggestion
            echo "! Failed to delete the githooks release repository" >&2
```

README.md
```diff
@@ -213,6 +213,15 @@ Optionally, you can also pass the template directory to which you want to instal
...
 ```
 
+When you want to install from another url (e.g. from your own fork) you can specify the release repository url 
+as well as the branch name (default: `master`) when installing with:
+
+```shell
+sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --release-clone-url "https://server.com/my-githooks-fork.git" --release-clone-url "release"
+```
+
+which will then be used for installation and further updates.
```
rycus86, line 223:
```suggestion
If you want to install from another repository (e.g. from your own fork), you can specify the release repository URL 
as well as the branch name (default: `master`) when installing with:

```shell
sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --release-clone-url "https://server.com/my-githooks-fork.git" --release-clone-branch "release"
```

This will be then used for installation and further updates.
```

File: README.md [hunk 27]
```diff
@@ -213,6 +213,15 @@ Optionally, you can also pass the template directory to which you want to instal
 $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --use-core-hookspath --template-dir /home/public/.githooks
 ```
 
+When you want to install from another url (e.g. from your own fork) you can specify the release repository url 
+as well as the branch name (default: `master`) when installing with:
```
rycus86, line 217:
```suggestion
If you want to install from another repository (e.g. from your own fork), you can specify the release repository URL
as well as the branch name (default: `master`) when installing with:
```

base-template.sh
gabyx: :+1:

README.md
```diff
@@ -213,6 +213,15 @@ Optionally, you can also pass the template directory to which you want to instal
...
+
+```shell
+sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --release-clone-url "https://server.com/my-githooks-fork.git" --release-clone-url "release"
```
rycus86, line 220:
```suggestion
sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --release-clone-url "https://server.com/my-githooks-fork.git" --release-clone-branch "release"
```

README.md
rycus86, line 223:
```suggestion
This will be then used for installation and further updates.
```

base-template.sh
gabyx: Hm.. jeah might be better. We anyway only pull if we dont have changes and are on the right branch, so it will work.

rycus86:
Thanks a lot for your work on this @gabyx !
I've added a bunch of comments, let me know what do you think! Thanks!

base-template.sh
gabyx: Correct. With git clone this is not so much a problem. But we had some strange errors because of this, better we guard this with a few extra options (unfortunately)

base-template.sh
gabyx: Resolve when rebasing

base-template.sh
gabyx: Rebasing

base-template.sh
gabyx: We could do that, might be less intrusive :-) :+1: Will test that.

base-template.sh
gabyx: Is this necessary? I think it should not be necessary, we use the variable always with quotes...?
I mean all over the code base we use `a=$(....)`. Or is there any significant difference? The assignmen is the same... the usage of `a` will depend on quotes or not...?

cli.sh
gabyx: Rebasing

install.sh
gabyx: This if is only entered when:
- we are not an update and
- we run the script from lets say `curl/wget` (externally not inside the release clone which is the ground truth)

install.sh
gabyx: Hm... we could do that of course. But this will only work if we always have a release clone: I think we have now? Will check that.

install.sh
gabyx: Rebasing

uninstall.sh
gabyx: :+1:

cli.sh
gabyx: Removed.

gabyx: Removed.

base-template.sh
gabyx: :+1:

tests/coverage.sh
gabyx: Oh, think it was a copy paste from exec-test.sh
Now its the same...

README.md
gabyx: :+1:

base-template.sh
gabyx: Hm. I think we should not add another indirect complication over a merge of the configured release branch to a possible changed (by the user) release branch. I think we just pull the configured branch, thats it :-). 
However I will introduce a warning, when we are dirty or are on some other branch (the user fiddled with it) we will abort the update with a warning you suggest :-)

gabyx: Ok. I come to the conclusion, that the behavior of completely re-clone if something is fishy is ok IMO. In this way we can also provide a 
```shell
git hooks config [set|print] release-clone-url
git hooks config [set|print] release-clone-branch
```
In the cli.sh. If the user changes this -> on the next update it will gracefully replace the existing clone which is nice.
I will however print a warnin when a pull fails (from whatever reason) and tell the user to error and a suggestion to manually update/fix or delete the release clone repo such that further update will succeed.
Is that ok?

rycus86: Yeah, that sounds good to me. This way they'd know that the local release directory shouldn't change, and hopefully would raise an issue here if they really want that. Otherwise their fork should work as expected if they're using a fork.

install.sh
gabyx: Please close this, if this answers your question.

install.sh
gabyx: Works.

gabyx: Hm.. tried this, and there would be some work to do because of the tests which do not work that way. I think its not so bad to copy the script to the /bin folder. since that is the install step, pull the release clone and install everythin from it.
I think in another PR we could further improve this. Reference `cli.sh` and possible `templates` (which is more cumbersome...)

rycus86: Yeah, I'm happy to do this later. In that case though, shouldn't we just `cp` over the CLI script instead of creating a new wrapper script?

gabyx:
Please check the latest commit, some rename and inclusion of some cli.sh funtionality.

base-template.sh
gabyx: Works.

gabyx:
One unresovled issue left, then we would be able to merge this :-) which I am looking forward too =).
**Note:** I would need to change the cli.sh dispathc in /bin to a copy still, this is better behavior for now...

@rycus86 : A suggestion: Would you welcome a PR (after this one) which basically gathers general functionality into `general.sh` which we can either source in cli.sh and also base-template.sh or directly include into (as before). In that way we would greatly improve the code basis since a lot of functionality is quadruplet :-) (install, base-template, uninstall. cli) etc... What do you think? With this PR things will get easier for such an approach

base-template.sh
rycus86: Nope, not necessary if spaces in those results don't break the assignment, if I remember correctly, they won't.

rycus86:
OK, thanks for the ping! I'll have a final pass over this in a bit, then it should be good to go.

> Would you welcome a PR (after this one) which basically gathers general functionality into general.sh which we can either source in cli.sh and also base-template.sh or directly include into (as before). In that way we would greatly improve the code basis since a lot of functionality is quadruplet :-) (install, base-template, uninstall. cli) etc...

Yep, agree. Now that we know that we have all the files from the repo locally, we should be able to include functions from those known script files, so yeah, totally open to it.

matthijskooijman:
I haven't reviewed this PR in detail, but there are two things I originally suggested in #83 that are not part of this PR yet. For ease of review, they are probably best done in a separate PR after this one, though.

 1. `install.sh` still copies the full `base-template.sh` content into the individual hooks. This leads to a lot of duplication of code (once per hook for a global template or global hook dir, once per hook per repo when installing into repositories) and makes it more work to update hooks. I would suggest installing not a copy of `base-template.sh`, but a symlink or tiny wrapper, so that there is always only one actual copy of `base-template.sh` present (which should maybe then also be renamed, since it will not longer be a template).
2. Individual hooks are still created by `install.sh`, rather than being pre-created in the repository. If they would be pre-created, installation could (also) be done without `install.sh` by simply setting the git global hooks directory to the directory inside the repo with the hooks (and combined with 1, this needs no real code duplication other than a single-line wrapper maybe).


Since symlinks are probably not portable to Windows, a one-line wrapper is probably better. For implementing that, there's two challenges:
 - Finding out where `base-template.sh` lives. For pre-created hook wrappers, this could probably do something like `DIR=$(cd $(basename $0) && pwd)` which is a common trick to get the absolute path to the current script even when symlinks or relatived paths are involved. The path to `base-template.sh` can then be derived from the hook dir. When installing the wrappers into the actual repository hook dirs, this won't work because `base-template.sh` lives somewhere else completely. In that case, you could hardcode the path in the wrapper script when copying it in `install.sh`, or maybe the wrapper could just read the install path from a git config variable (I suspect this is already set by `install.sh`?). The latter has the extra advantage that moving the git checkout is possibly without having to reprocess all git repos.
 - Executing `base-template.sh` without changing `$0` (so `base-template` still knows which hook was called). I think this can be done with plain `sh` using the `-c` option. E.g. `exec sh -c /path/to/base-template.sh "$0" "$@"`. Here, -c specifies the command to execute, while the arguments after that specify `$0`, `$1`, etc. Haven't tested this yet, though.

rycus86:
Thanks @matthijskooijman ! I agree these should be looked at in separate PR(s).

1. This is consistent with Git template dirs, which is the original way of Githooks was. The easier updates is an interesting aspect, I'd consider this for that only really, I'm not really convinced the duplication is necessarily bad here.

2. Isn't this essentially the same as the core.hooksPath installation method we already have?

matthijskooijman:
> 2. Isn't this essentially the same as the core.hooksPath installation method we already have?

No, since for that installation method, `install.sh` still has to *generate* a directory with all the hooks (with copies or wrappers of `base-template.sh`). What I'm suggesting is to pre-generate this directory and commit it (if they're thing wrappers, you'll have some code duplication but only a single-line script that should pretty much never need to be changed). With such a pregenerated, committed directory, you can just set `core.hooksPath` manually, without ever needing `install.sh`.

rycus86:
What's the benefit of that? I'm not sure I see the problem you're trying to solve?

matthijskooijman:
The "problem" of having to run an install script. If the hooks would be pregenerated, you could (optionally) install githooks by just cloning and setting a git config hook, without needing the install script at all.

Also, it would mean that there is no (generated) code anywhere which is *not* tracked by git. It seems unlikely that these tiny wrappers would need to be modified, but when they are git tracked, you can at least be sure (and if they *do* need a fix, updating them is just a matter of a git pull).

See also https://github.com/rycus86/githooks/issues/83#issuecomment-616623562

rycus86:
Yep, as mentioned there, for your use-case, the best solution would probably be what you described in https://github.com/rycus86/githooks/issues/83#issuecomment-614800948 - I don't really think though that your use-case is so common that it warrants a new installation method.

The whole promise of this project is that you run a single-liner install script, and you're ready to go. If that is not what you desire, it can be worked around it as you mentioned. I'm concerned about trying to support all the installation methods we can think of as some of them would work for some, and some for some others, but it could get confusing. If all you need to get where you want to be is to have those hook wrappers in a single directory wrapping a call to `base-template.sh`, we can probably get that added with a little blurb in the README. If you open a PR, we can perhaps discuss it more effectively?

README.md
```diff
@@ -213,6 +213,13 @@ Optionally, you can also pass the template directory to which you want to instal
 $ sh -c "$(curl -fsSL https://r.viktoradam.net/githooks)" -- --use-core-hookspath --template-dir /home/public/.githooks
 ```
 
+If you want to install from another repository (e.g. from your own fork), you can specify the update repository url 
+as well as the branch name (default: `master`) when installing with:
+```shell
```
rycus86, line 218: Missing new line before the code block and missing `$` before the shell command. I'll add these directly to master after merging.

base-template.sh
rycus86: I'll fix this up after merging. /after

cli.sh
```diff
@@ -1493,60 +1493,135 @@ call_script() {
...
+        GITHOOKS_CLONE_URL="https://github.com/rycus86/githooks.git"
+    fi
+
+    if [ -z "$GITHOOKS_CLONE_BRANCH" ]; then
+        GITHOOKS_CLONE_BRANCH="master"
+    fi
+
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "$INSTALL_DIR/release" >/dev/null 2>&1; then
+            echo "! Failed to remove an existing githook release repository" >&2
```
rycus86, line 1599:
```suggestion
            echo "! Failed to remove an existing githooks release repository" >&2
```

/after

cli.sh
rycus86: I'll fix this up. /after

cli.sh
```diff
@@ -2143,6 +2231,52 @@ config_update_state() {
...
 
+#####################################################
+# Manages the automatic update clone url.
+# Prints or modifies the
+#   \`githooks.autoupdate.updateCloneUrl\`
+#   global Git configuration.
+#####################################################
+config_update_clone_url() {
+    if [ "$1" = "print" ]; then
+        echo "Update clone url set to:"
```
rycus86, line 2242: This should be on a single line. /after

cli.sh
```diff
@@ -2143,6 +2231,52 @@ config_update_state() {
...
+
+#####################################################
+# Manages the automatic update clone branch.
+# Prints or modifies the
+#   \`githooks.autoupdate.updateCloneUrl\`
+#   global Git configuration.
+#####################################################
+config_update_clone_branch() {
+    if [ "$1" = "print" ]; then
+        echo "Update clone branch set to:"
```
rycus86, line 2265: This should be on a single line. /after

docs/command-line-tool.md
```diff
@@ -259,16 +273,17 @@ $ run <title> <text> <options> <long-options>
 ```
 
 The arguments for are:
+
 - `<title>` the title for the user gui dialog
 - `<text>` the text for the user gui dialog
-- `<short-options>` the button return values, slash-delimited,
```
rycus86, line 264: Not sure why these have changed, I'll revert it. /after

matthijskooijman:
> I don't really think though that your use-case is so common that it warrants a new installation method.

Yeah, I can see that. Though I guess this same pre-generated directory could also be used by the install script in the "global template dir" or "global hooks dir" scenarios (just not in the "install into repo hooks dir" scenario, because then the path to `base-template.sh` would be different).

>  we can probably get that added with a little blurb in the README. If you open a PR, we can perhaps discuss it more effectively?

That's fair, I'll see if I can find some time for that later. Regardless, I think that fixing 1. of https://github.com/rycus86/githooks/pull/86#issuecomment-619825822 would be separate from this (and probably best done before trying to fix or document 2.)?

install.sh
```diff
@@ -4049,8 +616,9 @@ setup_hook_templates() {
 ############################################################
 install_command_line_tool() {
     mkdir -p "$INSTALL_DIR/bin" &&
-        echo "$CLI_TOOL_CONTENT" >"$INSTALL_DIR/bin/githooks" &&
-        chmod +x "$INSTALL_DIR/bin/githooks" &&
+        echo "#!/bin/sh" >"$INSTALL_DIR/bin/githooks" &&
+        echo "sh '$INSTALL_DIR/release/cli.sh' \"\$@\"" >>"$INSTALL_DIR/bin/githooks"
+    chmod +x "$INSTALL_DIR/bin/githooks" &&
```
rycus86, line 621: The `chmod` should be chained with `&&` so we only do it if previous commands were successful, I'll fix it. /after

install.sh
```diff
@@ -4607,6 +1175,127 @@ set_githooks_directory() {
...
+
+    if [ "$PULL_ONLY" = "true" ]; then
+        PULL_OUTPUT=$(
+            git -C "$CLONE_DIR" \
+                --work-tree="$CLONE_DIR" \
+                --git-dir="$CLONE_DIR/.git" \
+                -c core.hooksPath=/dev/null pull origin "$GITHOOKS_CLONE_BRANCH" 2>&1
+        )
+
+        #shellcheck disable=2181
```
rycus86, line 1224: I'll do a replace of these. /after

install.sh
```diff
@@ -4607,6 +1175,127 @@ set_githooks_directory() {
...
+            -c core.hooksPath=/dev/null \
+            --depth 1 \
+            --single-branch \
+            --branch "$GITHOOKS_CLONE_BRANCH" \
+            "$GITHOOKS_CLONE_URL" "$INSTALL_DIR/release" 2>&1
+    )
+
+    #shellcheck disable=2181
+    if [ $? -ne 0 ]; then
+        echo "! Cloning of repository \`$GITHOOKS_CLONE_URL\`"
```
rycus86, line 1273: I'll make this single line. /after

tests/coverage.sh
```diff
@@ -16,7 +16,8 @@ RUN echo 'deb http://deb.debian.org/debian stretch main' >> /etc/apt/sources.lis
...
+ADD base-template.sh install.sh uninstall.sh cli.sh /var/lib/githooks/
```
rycus86, line 19: I'll change this back. /after

uninstall.sh
```diff
@@ -406,7 +406,22 @@ uninstall_shared_hooks() {
...
+
+############################################################
+# Uninstall the release repository.
+#
+# Returns:
+#   None
+############################################################
+uninstall_release_repo() {
+    if [ -d "$INSTALL_DIR/release" ]; then
+        if ! rm -rf "${INSTALL_DIR:?}/release" >/dev/null 2>&1; then
```
rycus86, line 423: I'll drop the `:?` here, we don't really use it elsewhere. /after

docs/command-line-tool.md
rycus86: Nope, this is how it should be, I'll leave it.

rycus86:
> > I don't really think though that your use-case is so common that it warrants a new installation method.
> 
> Yeah, I can see that. Though I guess this same pre-generated directory could also be used by the install script in the "global template dir" or "global hooks dir" scenarios (just not in the "install into repo hooks dir" scenario, because then the path to `base-template.sh` would be different).

For the existing installation methods, I'm happy with the copying those files around, pointing them back to the `base-template.sh` with an absolute path potentially.

> > we can probably get that added with a little blurb in the README. If you open a PR, we can perhaps discuss it more effectively?
> 
> That's fair, I'll see if I can find some time for that later. Regardless, I think that fixing 1. of [#86 (comment)](https://github.com/rycus86/githooks/pull/86#issuecomment-619825822) would be separate from this (and probably best done before trying to fix or document 2.)?

Yep, I agree.

rycus86:
Thanks a lot again @gabyx for your work on this! :tada:

tests/coverage.sh
gabyx: thx

docs/command-line-tool.md
gabyx: Was a markdown format in vs code :|

README.md
gabyx: thx

gabyx:
Ok, I try to wrap up things here as it got a bit complicated :-):

@matthijskooijman Would like to essentially place **wrappers** (simulating symlinks) (of course by using `git config githooks.installDir`) into the hooks forlder. That would be nice, but has at the moment some problems with 
- `--single` installs. These installs need a full-blown hooks script inside the hooks folder and not a **wrapper** (because they never get updated or only manually, `--single` is like "detached" from updates)
- using these **wrappers** gets a little bit messy, as we really need to distinguish between **wrappers** or full-blown scripts... (note: cli.sh and other stuff), I think its possible to handle both cases without any hassle, but thats only my guess...
- forgetting `--single` installs that would make things easier, but there is still  the **update disable** feature, such that you do not update the hooks inside a repo, so **wrappers** are implicity updated which we dont want...

So with the current features (leaving out the discussion about their usefulness) its kind of  troublesome. However I agree that having pregenerated **wrapper** hooks checked in would be lovely... but I think copying stuff isn't that of a problem (ok, we have duplicates for each hook script, which is a bit stupid....): But we could maybe do this:

## Method 1
**pregenerated hook wrapper (e.g pre-commit)**
```shell
DIR=$(cd $(basename $0) && pwd)
if is_not_single_install && is_update_disabled; then
# dispatch to
  $INSTALL_DIR=$(git config githooks.installDir)
  $INSTALL_DIR/release/base-template.sh ...
else
  # use template next to us...
  $DIR/githooks/base-template.sh  # LINE A (might source other files inside $DIR/githooks/)
fi
```
So we have a `githooks.sh` (base-template.sh) in the hooks folder by default for **any** install (no matter what) this is the only script which gets updated during update.
Any **non-single-install** repo and **non-update-disabled** repo will just use the release folder which will exist. 

A future PR, I mentioned, about `sourcing` general functionality is also not so easy because of `LINE A` above. So we would need to place all needed scripts also into the hooks folder. In that case this means we have a set of files `base-template.sh, etc...` which we could deploy to `.git/hooks/githooks` folder. If we would do that for every install into a repo, no matter what even if we dont use them and use the `$INSTALL_DIR`, we do **not really gain anything** (IMO) since these files `.git/hooks/githooks/*` need to be updated anyhow. So copying the `base-template.sh` for every hooks is not so far of the galaxy.

I think a valable solution would be to have the **wrappers** only supporting this:
## Method 2
**pregenerated hook wrapper (e.g pre-commit)**
```shell
  DIR=$(cd $(basename $0) && pwd)
  $DIR/githooks/base-template.sh
```
Then I would replace the name `base-template.sh` with just `githooks.sh` which is the main entry script for the manager...
I only think of such a PR for gathering shared functionality into scripts...
@rycus86 What do you think?

matthijskooijman:
## Single / disabled auto-update
> --single installs. These installs need a full-blown hooks script inside the hooks folder and not a wrapper (because they never get updated or only manually, --single is like "detached" from updates)

AFAIU, single is essentially the same as auto-update disabled, but additionally prevents the templates from being installed for future repos, right? For the purpose of this discussion, I think single and auto-update are essentially the same thing.

For new repos, disabled updates would be a matter of just copying `base-template.sh` into the git repo instead and using that local copy from the wrappers rather than the globally installed `base-template.sh`. In the future, this might also include other files that `base-template.sh` depends on.

For existing repos, things get a bit more complicated. Disabling automatic updates would mean retro-actively copying in `base-template.sh` (et al) into the git repo and updating the wrapper scripts to use it (which can happen through a git config variable).

### Needed at all?

I do wonder: Do we really need disabled auto-update at all? I can't quickly come up with any specific usescases for this. I can imagine that if updating is an explicit action, tracking repos and copying files around to them, you might want to not do that automatically (to stay in control). Also, because disabling auto-update is fairly easy to implement without adding complexity, that would make sense. But is it really needed?

Note that a single install, e.g. installing just some scripts into a single repo without making any git clone of this githooks repository elsewhere, I can imagine that that would still be supported, but that can just copy the needed files (either through a temporary clone, or maybe directly from git).

## Method 1
> If we would do that for every install into a repo, no matter what even if we dont use them and use the $INSTALL_DIR, we do not really gain anything (IMO) since these files .git/hooks/githooks/* need to be updated anyhow.

Yeah, *if* you want to keep supporting update-disabled exactly as now, by flipping a git config var, then you would indeed need to update these files even when unused, which is pointless and probably even more confusing for users.

## Method 2
So let me check if I got the idea right:
 - You have pre-generated wrapers that live in the githooks repo
 - For `core.hooksPath` installs, git would just be pointed at that directory, so updating is just a matter of git pull.
 - For in-repository installs, the entire `hooks` directory (wrappers and `base-template.sh`) can just be copied into the target git repository.
 - How would templating work? I originally thought that the template dir (`init.templateDir`) could simply be pointed at the pre-generated hooks directory, but looking more closely I see that this is a template for the entire `.git` directory, not just the hooks. So a user might already have a template directory and you would really need to install into (e.g. copy the pregenerated hooks directory into) that existing directory rather than just point git at the githooks repo.

## Method 3 (new)
Method 2 still needs the (I suspect somewhat complicated) auto-update logic (which tracks all git repositories and copies new files into them). I guess my ideal approach would simply not need this anymore at all, where you would either:
 - Copy actual code (not just wrappers) into the git repo and not get any automatic updates.
 - Copy wrappers into the repo (possibly via the template dir) that reference to githooks repo (possibly via a `githooks.installDir` config) so always get auto-updated.
 - Use `core.Hookspath` to point at (pre)generated hooks in the githooks repo, so always get auto-updated.

Switching between auto-updates and no auto-updates would be more tricky (essentially need a re-install into the repo, I guess).

Doing this would probably allow greatly simplifying updates (just do a git pull). 
OTOH, some of the auto-update copying code would still need to be kept around for backward compatibility, you'd still have to update existing repos to the new scheme, so maybe that voids these advantages...

## Wrapper script

Regardless of the method chose, I think the wrapper script should use the `.` operator to start the actual script:

      DIR=$(cd $(basename $0) && pwd)
      . $DIR/githooks/base-template.sh

Without the `.`, `$0` will be set to `base-template.sh`, losing info about what hook is called. I [previously suggested using `sh -c` for that](https://github.com/rycus86/githooks/pull/86#issuecomment-619825822), but just using `.` seems easier.

rycus86:
Yeah, I was thinking of something like Method 3 above. I'd like to see how this works on a PR, as it could get complicated a little bit, with trying to point both the Git template dir and the core.hooksPath to a set of pre-generated files, or pehaps we'd need multiple sets of pre-generated files.

Single install is a bit of a wildcard, I wouldn't worry too much about it. I'd copy the hooks into the single install repo as necessary, not pointing outside of it at all. Then if we want to make it not-single (if that's an option today), then we can re-run the installation as it was suggested above.

Auto-updates are a bit interesting, and I guess I could have thought about this before merging this PR. Previously, it would have checked if we have a new update and ask the user if they want to install it right now. With this PR, we'll always update I guess - do we even have the prompt still @gabyx ?
I guess we could always store the active commit hash and hard reset there if the user doesn't want to update just yet.
For disabling the auto-update, we can perhaps make that a global only option, then you won't be able to change this per repo, but I think that's probably fine.

For the wrapper, I don't really mind how it works as long as it works on POSIX and Git Bash on Windows.

rycus86:
Also, should we move this discussion back to #83 instead of this closed PR? 🙂

rycus86:
> With this PR, we'll always update I guess - do we even have the prompt still @gabyx ?

OK, looks like the prompt is still there but we will have updated the local clone already at that point. I guess it's not the end of the world. In the seamless-update mode we're planning, I guess users won't be given a prompt, we just display that we have updated if auto updates are enabled.

gabyx:
Thx: @matthijskooijman  For the summary. I think all your findings are correct, and your feeling about single repo install is the same I have. Its basically the same as disabling the auto update.

Meaning copy everything to the repo which is needed.

So can we agree on something like: 
- Make the **single** install equal to **disable update**
- Keep the registration mechanism for **update-disabled repos** (as it would be nice, since it helps the user to see all repos where updates are disabled, and there is a prompt asking if he would like to update also these update-disabled repos (maybe we could) just print these registered repos and -> let the user know "To update them please run `git hooks update --only-disabled`" which runs through all registered repos and copies the new hooks into them.

So the only change we have is that we basically place wrapper scripts, and when updates are disabled and enabled we do copying or wrapper-placement, thats all... 
I might have time to make PR and see how it goes.

rycus86:
From the README: 

> There is also an option to run the install script for the repository in the current directory only, without setting up the Git templates for any future repositories. For this, run the command below.
> ...
> You can change this setting later with the command line helper tool, running the git hooks config [set|reset] single command, which affects how future updates are run, when started from the local repository.

So yeah, it's mostly about disabling update, but also not installing anything outside of the current repo you're installing into.

To be honest, if we're going to do wrappers pointing back to a central checkout of the base template, I'd say let's just deprecate single install very quickly, then let's break it when the wrappers are ready to go. It will be a nightmare having to maintain it and switch back and forth between single and regular installs too. If you can think of a way to keep it working and not be a burden, then we can talk about it though.

For the same reason, I would remove the repo tracking as well, just have a global switch for updates or no updates, then at an actual update we just do a git pull of the target branch and that's it (possibly some post-install script in case we do need to set something up).

How does this sound to you both?

gabyx:
Oh, jeah, I tried to keep it working in my test changes and its really a nightmare (also to maintain and understand the logic. I think deprecating this feature is the best option.

Thanks for your suggestion, it is very much in line with my thoughts. Its probably the best, and things get easier! I will try this tonight. Will simplyfy a lot I think. 
Only having a global update is better.

rycus86:
All right, looking forward to the PR!

We should also get the deprecation started, would be good to have some warning showing up for single install repos for at least a week before it disappears, though not super certain that's really that much time... At the same time, no idea if people do actually use single install or not, having no analytics on things. :(

I'll see if I can get a change in this week to add a warning message.

matthijskooijman:
Yeah, simple sounds good :-D

I guess you would still have three flavours:
 1. Install wrappers into separate repositories
 2. Install wrappers into a template directory
 3. Install wrappers into a global hooks directory (or point the global hooks directory at a pregenerated global hooks directory in the repo, as discussed before).

All of these would point to the same base script that lives in a git checkout, so all of these update in the same way. Right?

As for migration - I guess you could maybe do a one-off conversion of all repos to the new wrapper-based install based on the existing repo tracking (maybe ask for confirmation first and cancel the update otherwise)?

gabyx:
See #92

PR #88, opened 2020-04-23
Test if /dev/tty can be opened before reading it 

gabyx:
Better implementation as suggested from #85. 
The prompt needs to be shown because of the tests and I would leave it like this...

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2004.170929-8e4d56
+# Version: 2004.251306-8304f2
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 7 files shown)


base-template.sh
```diff
@@ -760,15 +760,17 @@ show_prompt() {
...
-    # from git), so read from /dev/tty, our controlling terminal.
-    # However, only do this when stdout *is* a tty, otherwise it is
-    # likely we have no controlling terminal and reading from /dev/tty
-    # would fail with an error.
+    # from git), so read from /dev/tty, our controlling terminal,
+    # if it can be opened.
     printf "%s %s [%s]:" "$TEXT" "$HINT_TEXT" "$SHORT_OPTIONS"
-    # shellcheck disable=SC2229
-    read -r "$VARIABLE" </dev/tty 2>/dev/null
+    #shellcheck disable=2217
```
rycus86, line 766:
```suggestion
    # shellcheck disable=SC2217
```

For consistency.

rycus86: Looks good otherwise, thanks @gabyx !
Could you please apply the suggestion, then I'll merge on the green Travis build. Thanks!

gabyx: :+1:  Done. Regex replaced all of them which where not as you wish.

rycus86:
Thanks a lot!

PR #90, opened 2020-04-27
Bugfix: guard run_update_check

gabyx:
Small CLI fix:
Function `run_update_check` -> `read_single_repo_info` might be executed  while not in a repository, outputting an error.

Thx for merging.


rycus86:
Thanks!

PR #93, opened 2020-05-18
Make single installs deprecated.

gabyx:
This is PR Nr.1 towards making single installs completely disappear.
In this PR **only**:

-  a little bit of clean up and refactoring:
    - make "true,no,yes" stuff only "true/false"
    - no env vars, directly use arguments to scripts
- `cli.sh` is directly referenced to the `$INSTALL_DIR/release` folder.
- solely `install.sh` is responsible to merge changes into the local branch, `base-template.sh` and `cli.sh` only asserts that a clone needs to be there. Therefore we improved the update logic which is gonna be needed in a further PR to introduce the base-template symlink and legacy clean up stuff.
- Deprecation warning (`test-039`) for installs into single repos.  
- Registration mechanism is left as is, registration is done for all repos (good feature for future stuff!).


gabyx:
Once this is reviewed **not merged**, I continue to incorporate in another PR the changes needed to symlink the `base-template.sh`. So we need to wait till all PR's Nr1 till the end are reviewed. And merge them together into the master.
@rycus86 : thanks for the review.

gabyx:
I dont know why 2 tests in `test-arch` fail. Locally they dont. Its the same shitty problem...?

rycus86:
I'm a bit confused with this PR again, from the title I thought this is going to be the first step: deprecating single installs as best as we can, failing new single installs, then that should be merged.

This PR also seems to change updates, versioning, and a few.other things as well. Can we please have one change per one PR for now? That might also help narrowing down the problem with the Arch tests as well.

gabyx:
The problem is, these changes we make in *your* envisioned PR will all be reverted and refactored again after we have another PR, For my point of view this is very cumbersome and tedious. Look at the two commits. The last one is only docker related (thats why 26 files...). I will try to norrow down the arch problems later this day maybe. With this PR its a tradeoff between changes and the next changes we are going to to -> wrapping the base-template.
Please have a look.

gabyx:
This changes should be mergable yes.

rycus86:
> The problem is, these changes we make in your envisioned PR will all be reverted and refactored again after we have another PR

Yes, I know, but I wanted to give people a chance to update to the version with the warning before removing support altogether. I'm happy to work on a PR that does only that, but it will be a couple of days at least before I'll have time to look into that most likely.

gabyx:
Aha, ok. Sorry there was a missunderstanding, then. I will try to incorporate only such a change. 
Do you want a PR which shows a warning, and the single install feature still continues to work, but telling the user that he/she should switch single off...?

rycus86:
That's OK, sorry if I didn't explain myself properly. I think what we could perhaps do is, print a warning and fail the installs with the single flag, and also warn and fail updating single install repos. I think that's all we can do right? We don't change single install repos in any other cases, so this is probably all we can do to warn users.

PR #94, opened 2020-05-19
Deprecation warning added for single installs

gabyx:
Only a warning, to let user know that single installs are deprecated with further installs.

No changes to any logic. Documentation marks all places as deprecated.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ env:
   - TEST_INSTANCE=test-debian-user
   - TEST_INSTANCE=test-ubuntu
   - TEST_INSTANCE=test-centos
-  - TEST_INSTANCE=test-arch
+  - TEST_INSTANCE=test-arch # Test 103/109 have permission problems -> travis bug
   - TEST_INSTANCE=test-fedora
   - TEST_INSTANCE=test-opensuse
   - TEST_INSTANCE=test-corehookspath
--- a/docs/command-line-tool.md
+++ b/docs/command-line-tool.md
@@ -173,7 +173,7 @@ Disables running any Githooks files in the current repository, when the `set` op
 $ git hooks config [set|reset|print] single
 ```
 
-Marks the current local repository to be managed as a single Githooks installation, or clears the marker, with `set` and `reset` respectively. The `print` option outputs the current setting of it. This command needs to be run at the root of a repository.
+**[deprecated]**: Marks the current local repository to be managed as a single Githooks installation, or clears the marker, with `set` and `reset` respectively. The `print` option outputs the current setting of it. This command needs to be run at the root of a repository.
 
 ```shell
 $ git hooks config set search-dir <path>
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 2006.021431-5f4f28
+# Version: 2006.022309-09ca86
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -88,6 +88,10 @@ execute_installation() {
         setup_shared_hook_repositories
         echo # For visual separation
     fi
+
+    thank_you
+
+    return 0
 }
 
 ############################################################
@@ -194,11 +198,63 @@ parse_command_line_arguments() {
         prev_p="$p"
     done
 
+    if [ "$SINGLE_REPO_INSTALL" = "yes" ]; then
+        if ! check_not_deprecated_single_install; then
+            echo "! Install failed due to deprecated single install" >&2
+            exit 1
+        else
+            warn_deprecated_single_install
+        fi
+    fi
+
     # Using core.hooksPath implies it applies to all repo's
     if [ "$SINGLE_REPO_INSTALL" = "yes" ] && [ "$USE_CORE_HOOKSPATH" = "yes" ]; then
         echo "! Cannot use --single and --use-core-hookspath together" >&2
         exit 1
     fi
+}
+
+############################################################
+# Install flag `--single` will behave differently in
+# future versions.
+#
+# Returns: None
+############################################################
+warn_deprecated_single_install() {
+    echo "" >&2
+    echo "! FEATURE CHANGE WARNING: The single installation feature" >&2
+    echo "  with \`--single\` will be changed to the following only" >&2
+    echo "  bahavior in future updates:" >&2
+    echo "" >&2
+    echo "    - install Githooks hooks into the current repository" >&2
+    echo "    - the installed hooks are not standalone anymore" >&2
+    echo "      and behave exactly the same as current non-single" >&2
+    echo "      installs" >&2
+    echo "" >&2
+}
+
+############################################################
+# Check if this repo is not a deprecated single install.
+#
+# Returns: 1 if deprecated single install, 0 otherwise
+############################################################
+check_not_deprecated_single_install() {
+    if is_git_repo "$(pwd)" && git config --local githooks.single.install; then
+        echo "" >&2
+        echo "! DEPRECATION WARNING: Single install repositories will be " >&2
+        echo "  completely deprecated in future updates:" >&2
+        echo "" >&2
+        echo "    You appear to have setup this repo as a single install." >&2
+        echo "    The hooks will still work with this install but will not be" >&2
+        echo "    supported anymore in the next update." >&2
+        echo "" >&2
+        echo "    You need to reset this option by running" >&2
+        echo "      \`git hooks config reset single\`" >&2
+        echo "    in order to use this repository with the next updates!" >&2
+        echo "" >&2
+        return 1 # DeprecateSingleInstall
+    fi
+    return 0
 }
 
 ############################################################
@@ -1354,4 +1410,3 @@ thank_you() {
 
 # Start the installation process
 execute_installation "$@" || exit 1
-thank_you
--- a/install.sh
+++ b/install.sh
@@ -4,7 +4,7 @@
 #   and performs some optional setup for existing repositories.
 #   See the documentation in the project README for more information.
 #
-# Version: 2006.022309-09ca86
+# Version: 2006.022313-8dd877
 
 # The list of hooks we can manage with this script
 MANAGED_HOOK_NAMES="
@@ -224,7 +224,7 @@ warn_deprecated_single_install() {
     echo "" >&2
     echo "! FEATURE CHANGE WARNING: The single installation feature" >&2
     echo "  with \`--single\` will be changed to the following only" >&2
-    echo "  bahavior in future updates:" >&2
+    echo "  behavior in future updates:" >&2
     echo "" >&2
     echo "    - install Githooks hooks into the current repository" >&2
     echo "    - the installed hooks are not standalone anymore" >&2
--- a/tests/step-039.sh
+++ b/tests/step-039.sh
@@ -12,6 +12,14 @@ mkdir -p /tmp/start/dir && cd /tmp/start/dir || exit 1
 mkdir -p /tmp/empty &&
     GIT_TEMPLATE_DIR=/tmp/empty git init || exit 1
 
+git config --local githooks.single.install "Y"
+OUT=$(sh /var/lib/githooks/install.sh --single 2>&1)
+if ! echo "$OUT" | grep -iq "DEPRECATION WARNING: Single install"; then
+    echo "! Expected installation to fail because of single install flag: $OUT"
+    exit 1
+fi
+git config --local --unset githooks.single.install
+
 if ! sh /var/lib/githooks/install.sh --single; then
     echo "! Installation failed"
     exit 1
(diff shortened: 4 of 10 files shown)


rycus86:
Thanks a lot! This looks 👌 I'll wait for Travis then this should be good to go.

gabyx:
Jup, I will fix it. Still some errors, from cherrypick merge etc...

gabyx:
Had some f**** up error in arch tests -> the last `thank_you` is not executed, I could not find out why. 
It should have been executed. I took it inside the `execute_install` function now everything works... Might be a replacement error or something else weird...
```shell
execute_installation "$@" || exit 1
echo "A"
echo "B"
echo "C"
```
only B and C were printed, the `A` was never printed, also the thank_you which was there was never executed...??

gabyx:
Trying to still fix the weird arch tests 103 and 109 (something with permissions and `ls -A`) ...

gabyx:
Ok I cannot fix these arch tests 103/109, I dont know also a workaround, locally it works. It seems its travis and its docker setup ? ...

gabyx:
Can we anyway merge this change?

rycus86:
Can you try pulling the latest Arch image, we should be able to reproduce these issues locally, and I believe Travis will always rebuild from the latest image available on Docker Hub, while locally you might use a cached base image.

gabyx:
Checkd with the latest image. Locally no errors.

gabyx:
So what shall we do, its a travis problem...

gabyx:
I would remove the arch test. from travis until the stuff is fixed. we can of course check the arch tes locally until then...

rycus86:
Just re-ran the Arch test on master kn Travis and it's failing the same way, so happy to exclude it for now. Is this PR ready from your side?

install.sh
```diff
@@ -88,6 +88,10 @@ execute_installation() {
         setup_shared_hook_repositories
         echo # For visual separation
     fi
+
+    thank_you
```
rycus86, line 92: Why did this have to move here?

tests/step-039.sh
```diff
@@ -12,6 +12,14 @@ mkdir -p /tmp/start/dir && cd /tmp/start/dir || exit 1
...
 
+git config --local githooks.single.install "Y"
+OUT=$(sh /var/lib/githooks/install.sh --single 2>&1)
+if ! echo "$OUT" | grep -iq "DEPRECATION WARNING: Single install"; then
+    echo "! Expected installation to fail because of single install flag: $OUT"
+    exit 1
+fi
+git config --local --unset githooks.single.install
+
 if ! sh /var/lib/githooks/install.sh --single; then
```
rycus86, line 23: This should also fail, regardless of the `githooks.single.install` setting.
Once this PR is in, you can only update (e.g. run the install) if it's not single anymore.

tests/step-103.sh
```diff
@@ -22,16 +22,21 @@ mkdir -p .githooks && echo '/tmp/shared/hooks-103.git' >.githooks/.shared || exi
 git add .githooks/.shared
 git hooks shared update
 
-RESULT=$(ls -A ~/.githooks/shared 2>/dev/null)
+RESULT=$(ls -A ~/.githooks/shared 2>&1)
```
rycus86, line 25: Should we revert the debug messages in this and in 109?

.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
   - TEST_INSTANCE=test-debian-user
   - TEST_INSTANCE=test-ubuntu
   - TEST_INSTANCE=test-centos
-  - TEST_INSTANCE=test-arch
+  - # TEST_INSTANCE=test-arch Test 103/109 have permission problems -> travis bug
```
rycus86, line 18:
```suggestion
  # - TEST_INSTANCE=test-arch Test 103/109 have permission problems -> travis bug
```

docs/command-line-tool.md
```diff
@@ -173,7 +173,7 @@ Disables running any Githooks files in the current repository, when the `set` op
 $ git hooks config [set|reset|print] single
 ```
 
-Marks the current local repository to be managed as a single Githooks installation, or clears the marker, with `set` and `reset` respectively. The `print` option outputs the current setting of it. This command needs to be run at the root of a repository.
+**[deperecated]**: Marks the current local repository to be managed as a single Githooks installation, or clears the marker, with `set` and `reset` respectively. The `print` option outputs the current setting of it. This command needs to be run at the root of a repository.
```
rycus86, line 176:
```suggestion
**[deprecated]**: Marks the current local repository to be managed as a single Githooks installation, or clears the marker, with `set` and `reset` respectively. The `print` option outputs the current setting of it. This command needs to be run at the root of a repository.
```

install.sh
rycus86: @gabyx ?

install.sh
```diff
@@ -201,13 +205,47 @@ parse_command_line_arguments() {
...
     fi
 }
 
+############################################################
+# Check if this repo is not a deprecated single install.
+#
+# Returns: 1 if deprecated single install, 0 otherwise
+############################################################
+check_not_deprecated_single_install() {
+    if is_git_repo "$(pwd)" && git config --local githooks.single.install; then
```
rycus86, line 226: As mentioned on the test below, we should fail single installs everywhere, not only in already single install repos.
@gabyx can you please change the logic like that?

tests/step-103.sh
```diff
@@ -78,9 +83,9 @@ echo "Cloning"
 cd /tmp || exit 1
 git clone /tmp/test103 test103-clone && cd test103-clone || exit 1
 
-RESULT=$(ls -A ~/.githooks/shared 2>/dev/null)
+RESULT=$(ls -A ~/.githooks/shared 2>&1)
```
rycus86, line 86: I have a feeling the `ls -A` is breaking on Arch as that's not a POSIX flag by the looks of it.
I haven't had a chance to test it though, so I could be wrong.

rycus86:
@gabyx left some comments for you, sorry about the delays!
It looks mostly OK, just two things I want to highlight:
- single installs should now all fail, regardless where/how you're trying to run them
- the warning message can just say they're deprecated now and need to clear the setting before trying again, I'd drop the bit about "how this happened" not sure that adds much value to the users

gabyx:
- single installs should now all fail, regardless where/how you're trying to run them
I thought the following as already answered above:
> --single will become the normal "install only to this repo" command.
Up to now there is an **additional** "... and also make it standalone". So this test will still work and --single is an option which is still useful,
Its just not a **global** install asking for other stuff like "search other repos/install into registered repos..."

**Also** It can't be remove if we want `cd repo && git hooks install`. Since that supports this.

rycus86:
> --single will become the normal "install only to this repo" command

Yep, this will be equivalent to the normal install as single installs will eventually just point to the same base template all installations do. Perhaps we could drop the `--global` flag from that command and just run a non-single install.

> Also It can't be remove if we want cd repo && git hooks install. Since that supports this.
I think this is fine? Isn't it?

Now that I write this, I'm thinking just failing a single install on running `git hooks install` is fine, the user gets the error, then we'll make that flag disappear in a (near) future change.

gabyx:
Ok. Inputs for a near future PR on to this one:
The use case for `--single` is still
```shell
cd repo
git hooks install
```
why should that run a global installation, especially asking for 
- global shared hooks 
- asking to search for existing repos
- etc 
I would leave the --single flag for `install.sh` in.
And also dont change the `cli.sh`, isn't it good as it is?

rycus86:
Ah yeah, good point. So `git hooks install` should still work but shouldn't try to scan for template dirs or offer to set up shared hooks etc. If that's currently done with the `--single` flag on the install script, I'd change that so that flag can actually just error out with a message that helps the user understand what they need to do. The bit I'm not sure of what happens is when you can `git hooks install` using an old cli, will that fetch the latest install script and call it with `--single`? If that's the case then that won't work. If not with this PR then with one in the very near future. So perhaps we just need a nice error message to tell the user what they need to do and let it break. What do you think?

gabyx:
Exactly: vgit hooks install` -> Thats why the whole stuff gets complicated. It fetches the new install.sh and installs an update. Thats why we leave the `--single`  flag on `install.sh`. It makes perfect sense. 
**If the repo happens to be  a "single install" == meaning its standalone (`githooks.single.install` as I call it** -> we error out with a nice error message in this PR.

Makes that sense? Could we merge in #71 before (are just some small bug fixes)

gabyx:
Lets test again if the now POSIX compliant tests still fail in ARch

gabyx:
OK seems good to go. Thanks for merging, then I can go on with more changes...
Maybe; Could we merge in #71 before (are just some small bug fixes)

.travis.yml
```diff
@@ -15,7 +15,7 @@ env:
   - TEST_INSTANCE=test-debian-user
   - TEST_INSTANCE=test-ubuntu
   - TEST_INSTANCE=test-centos
-  - TEST_INSTANCE=test-arch
+  - TEST_INSTANCE=test-arch # Test 103/109 have permission problems -> travis bug
```
rycus86, line 18: If this is fixed now, could you please remove the comment?
```suggestion
  - TEST_INSTANCE=test-arch
```

rycus86:
I think I get your point now about keeping the `--single` to keep `git hooks install` working, but I'm not clear on the implementation here. We're saying that `git hooks install` should keep working in any case, right? So your changes would effectively only fail updates in repos that are already single installs?
What about new single installs, how do we let them know that this is changing soon? Also, perhaps we should have a message (or an error even) for `git hooks config single` or whatever that's called, so people wouldn't want to change to single repos from existing ones.
What do you think?

gabyx:
Yes thra would make lots of sense . I will add another depr. warning for this case!

Von meinem iPhone gesendet

gabyx:
Yes your findings are correct. Will only fail for single install repos with this flag...

Von meinem iPhone gesendet

tests/step-039.sh
```diff
@@ -12,6 +12,14 @@ mkdir -p /tmp/start/dir && cd /tmp/start/dir || exit 1
 mkdir -p /tmp/empty &&
     GIT_TEMPLATE_DIR=/tmp/empty git init || exit 1
 
+git config --local githooks.single.install "Y"
```
rycus86, line 15: I think perhaps leave this test to verify that single installs still work, but perhaps check the deprecation warming was present in the output.
I would then change test 042 to test that the update fails in a single repo install.
What do you think?

rycus86:
Thanks a lot, starting to look good!
So to clarify:
- new single installs will show the warning but successfully install as single repo
- updating single install repos will show a warning and fail
- changing single config prints the warning but still works

Did I get these right?

gabyx:
> Thanks a lot, starting to look good!
> So to clarify:
> 
> * new single installs will show the warning but successfully install as single repo

yes (hm.. we do this even if its not updatable anymore... but its for the PR's reason, keep it small)

> * updating single install repos will show a warning and fail

yes

> * changing single config prints the warning but still works

no
-> there should also be a warning, will add this! hm..

Thx for the summary.

tests/step-039.sh
gabyx: The  test039.sh of this PR actually tests 
- failing of the install
- single install should still work.

I will add the same in test 042. first checking that the update fails, then  unset the single flag  and
run the update.

.travis.yml
rycus86: Can you please change this @gabyx ?

install.sh
```diff
@@ -194,13 +198,65 @@ parse_command_line_arguments() {
...
+# Install flag `--single` will behave differently in
+# future versions.
+#
+# Returns: None
+############################################################
+warn_deprecated_single_install() {
+    echo "" >&2
+    echo "! FEATURE CHANGE WARNING: The single installation feature" >&2
+    echo "  with \`--single\` will be changed to the following only" >&2
+    echo "  bahavior in future updates:" >&2
```
rycus86, line 227:
```suggestion
    echo "  behavior in future updates:" >&2
```
Small typo here

rycus86:
Nice one! I think this is good to go now!
Could you please fix up the comment on the `.travis.yml` and the typo in the message shown to the user?
I'll merge it after that.

Thanks a lot for pushing this through!

rycus86:
As a side note, why are the Arch tests still failing?
I thought changing the `ls -A` fixed them.

gabyx:
No it did not I made it Posix compl. but that didnt help. its still a bug ...

Von meinem iPhone gesendet

rycus86:
Awesome, thanks a lot for your work on this PR! 👍

gabyx:
It will continue, sadly. But now we have a base to work on I think... Thx for mergin. I will refresh the old PR and continue with the deprecation. 
1. Call cli.sh from the release folder
2. Make an upgrade procedure which is safe and tries the best but doesnt clutter code hopefully. 
    Maybe we can outsource this upgrade into another script such that its fully independent, lets see...

rycus86:
Sounds good! Let's just make sure we go with small PRs so it's easy to verify they do what they should. 👌

gabyx:
Ok :-)

gabyx:
Just fyi: The problem with small PRs in this case is, that as soon as we **really** take single install off the place we need to fix all together in a single PR, because its a cobweb, I dont think making multiple PRs will help the problem, so you have to jump from one to the other?, I can however make single commits in one PR which can be digested better. I will see how I start...

rycus86:
It all depends on what is in scope of the PR. If it only changes one thing and one thing only, it can stay small.

PR #96, opened 2020-06-02
Bugfix: Create template directory in all cases :tada:

gabyx:
Szenenario: Fresh install
- set a (init.templateDir)  e.g `~/mytemplates`
- install fails with (`Git hook templates directory not found`) because mkdir is at the wrong place. 

See small Bugfix. 

Thanks for merging.


rycus86:
Thanks for the bugfix!

PR #97, opened 2020-06-02
Small refactor: logic after read wrong :zap:

gabyx:
Logic after read is potentially harmful.
Small refactor.

Thx for mergin.


gabyx:
Just a moment its wrong!

gabyx:
Ok now it should be correct

PR #98, opened 2020-06-03
Bugfix: Quote hook alias :boom:

gabyx:
- whitespace problem

Thx for merging

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2006.022313-8dd877
+# Version: 2006.031558-97a441
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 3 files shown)


rycus86:
Thanks for this! Is this a problem you've seen? We have a whitespace test on Travis, I thought that should have picked it up already?

gabyx:
Jeah on windows. We have these tests, but unfortunately not inside the `~` home folder...
had this on windows...

rycus86:
I like your suggestion, we should try that. Travis seems to be OK with this change on Linux, so let me merge it.

gabyx:
I have some more test changes ;-) for the `test-whitespace.sh`
Will open this in another PR, this will propery test this PR.

to what will `~` expand if it has whitespace inside?

gabyx:
We should prohibit install into a `~` which has whitespaces. It just gives lots of troubles...
or can we really handle this properly ?

PR #100, opened 2020-06-04
2: Refactoring all internal booleans :star:

gabyx:
- No arguments over environment anymore
  Backward compatible for update!
- Change all "yes" -> "true" to make it more git config like `git config --bool` and consistent. 

4 Commits changing variables...

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2006.051449-e3ae41
+# Version: 2006.051457-8e0144
 
 #####################################################
 # Execute the current hook,
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2006.051422-382787
+# Version: 2006.051442-decd77
 
 #####################################################
 # Execute the current hook,
@@ -682,7 +682,8 @@ should_run_update_checks() {
     [ "$HOOK_NAME" != "post-commit" ] && return 1
 
     UPDATES_ENABLED=$(git config --get githooks.autoupdate.enabled)
-    [ "$UPDATES_ENABLED" != "true" ] && return 1
+    [ "$UPDATES_ENABLED" != "true" ] &&
+        [ "$UPDATES_ENABLED" != "Y" ] && return 1
 
     CURRENT_TIME=$(date +%s)
     ELAPSED_TIME=$((CURRENT_TIME - LAST_UPDATE))
--- a/base-template.sh
+++ b/base-template.sh
@@ -4,7 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
-# Version: 2006.051420-807309
+# Version: 2006.051422-382787
 
 #####################################################
 # Execute the current hook,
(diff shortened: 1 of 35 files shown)


gabyx:
Thx for merging.

gabyx:
Can you have a look at this. This would be good to go.

rycus86:
How are the `Y` -> `true` and `N` -> `false` changes backwards compatible? I think they would just be invalid value and probably be handled on the fallback branches of the conditions.
Could you perhaps revert all the test changes and see that everything still works? That way we'd have better confidence that these changes are actually backwards compatible.
Thanks!

gabyx:
Ah jeah that might be a good idea,

gabyx:
The changes are backward compatible (IMO) because all tests are adapted such that they still test the old values
```shell
[ "$A" = "true"] || [ "$A" = "yes" ] || return 1
```

install.sh
```diff
@@ -709,7 +723,7 @@ install_command_line_tool() {
 ############################################################
 setup_automatic_update_checks() {
     if CURRENT_SETTING=$(git config --get githooks.autoupdate.enabled); then
-        if [ "$CURRENT_SETTING" = "Y" ]; then
+        if [ "$CURRENT_SETTING" = "true" ]; then
```
rycus86, line 726: This test for example would probably not consider auto updates to be enabled after this change, since it reads the config and will ignore the previous `Y`.
Again, I think reverting the test changes should surface these nicely to fix

gabyx: ah correct, there are some errors from my side!

cli.sh
```diff
@@ -2260,14 +2262,14 @@ config_trust_all_hooks() {
...
+        git config --global githooks.autoupdate.enabled true
     elif [ "$1" = "disable" ]; then
-        git config --global githooks.autoupdate.enabled N
+        git config --global githooks.autoupdate.enabled false
     elif [ "$1" = "reset" ]; then
         git config --global --unset githooks.autoupdate.enabled
     elif [ "$1" = "print" ]; then
         CONFIG_UPDATE_ENABLED=$(git config --get githooks.autoupdate.enabled)
-        if [ "$CONFIG_UPDATE_ENABLED" = "Y" ]; then
+        if [ "$CONFIG_UPDATE_ENABLED" = "true" ]; then
```
gabyx, line 2272: Here the same

base-template.sh
```diff
@@ -680,7 +682,7 @@ should_run_update_checks() {
     [ "$HOOK_NAME" != "post-commit" ] && return 1
 
     UPDATES_ENABLED=$(git config --get githooks.autoupdate.enabled)
-    [ "$UPDATES_ENABLED" != "Y" ] && return 1
+    [ "$UPDATES_ENABLED" != "true" ] && return 1
```
gabyx, line 685: Here too

gabyx:
Pushed the necessary tests to check

gabyx:
If all tests pass I will add the more changes to all tests to update them.

gabyx:
All tests in : https://travis-ci.org/github/rycus86/githooks/builds/695011929
work. Meaning backward compatibility should be :crossed_fingers: ok.

I will add the other tests changes back in.

gabyx:
Lets see if all tests pass now, if so, Its good to go :+1:

gabyx:
A next PR would be:
- cli.sh from release clone
or
- remove the nasty version number update in the files (always gives merge conflicts). should work with installs from very old versions... (we leave the version number in, for those who read this number), its complicated -> maybe later ...

gabyx:
Ok, everythin works except arch of course  -> Good to go :+1:

rycus86:
Thanks for this! 👌
For what's next:
> cli.sh from release clone
Could we please do this first to see how that goes?

> remove the nasty version number update in the files (always gives merge conflicts). should work with installs from very old versions... (we leave the version number in, for those who read this number), its complicated -> maybe later ...
I kind of like that version number to be checked in and available in the local file. I agree the merge conflict is somewhat annoying, perhaps we could work something out to keep the version number but make it less annoying? Any ideas?

PR #104, opened 2020-06-05
Make cli.sh work directly from the release clone

gabyx:


gabyx:
I am waiting for a response on this question:
https://stackoverflow.com/questions/62218749/does-git-create-new-files-during-merge-or-overwrites-existing-ones

gabyx:
Seems to work https://travis-ci.org/github/rycus86/githooks/builds/695079360
Good to go :+1:

gabyx:
Thanks for mergin

gabyx:
@rycus86 : Wait...
I am unsure if we really want to go down the road of having files executed in the release folder (which is updateable at the same time) because we rely on git that it really creates a new inode (which is the case)...

gabyx:
We could as a solution maybe before any update happens
copy the install.sh to a temporary location and execxute from there exactly at this line:
https://github.com/rycus86/githooks/blob/bf018cb078ae651667d26f28404ab80bb71e77a6/install.sh#L41

gabyx:
For me its good to go :+1:

rycus86:
This is conflicted, otherwise looks good to go.

tests/step-094.sh
```diff
@@ -51,12 +51,12 @@ if ! sh /var/lib/githooks/install.sh; then
...
-cp /tmp/cli-0 "$HOME/.githooks/bin/githooks" &&
-    chmod +x "$HOME/.githooks/bin/githooks" || exit 1
-
-if ! git hooks install; then
-    echo "! The Git alias integration failed: single"
-    # exit 1
+cp /tmp/cli-0 "$HOME/.githooks/release/cli.sh" || exit 1
+OUT=$(git hooks install 2>&1)
+# shellcheck disable=SC2181
+if [ $? -eq 0 ] || ! echo "$OUT" | grep -iq "DEPRECATION WARNING: Single install"; then
```
rycus86, line 57: This used to test that the `git hooks install` command works, but now we test that it fails. Perhaps we should unset the single repo is that's why this has changed?

gabyx:
Rebased corrected test094
+ as a bonus: 
Capped to long lines in echos to stdout, to make it more readable

gabyx:
All test pass > Good to go :+1:

rycus86:
> Capped to long lines in echos to stdout, to make it more readable

I think this changes some of the indentation logic that was used elsewhere (line continuation to be indented if I remember correctly), but OK to look at that later maybe.

Looks good to me, happy to merge once the conflict is resolved.

rycus86:
> Capped to long lines in echos to stdout, to make it more readable

I think this changes some of the indentation logic that was used elsewhere (line continuation to be indented if I remember correctly), but OK to look at that later maybe.

Looks good to me, happy to merge once the conflict is resolved.

gabyx:
Ah is that the style? I just seen this only a couple of times ? all other warnings and such dont use special line contin. indentation
Shit still a conflict , I rebased hmmm. Lets see...

Von meinem iPhone gesendet

gabyx:
Rebased to master

PR #105, opened 2020-06-05
1: Bugfix: Show deprecation warning only once

gabyx:
- small refactor for better understanding
- outsource warning and checks into check_deprecation
- explicit return values

Thanks for mergin


gabyx:
Its good to go :+1: 
https://travis-ci.org/github/rycus86/githooks/builds/695126986

gabyx:
I think I need to lool closer to the update from previous version since we have git now I can directly enforce this easily in a test

PR #106, opened 2020-06-05
2: Refactor: Offload update to install script

gabyx:
Continuation on #105 

- Move the whole update logic to the `install.sh`
  The base-template and cli.sh is only fetching the changes, the install merges the changes afterwards
- Version number is not used anymore only for printin in the console
- No other **logic changes** such as changes to single install stuff...
- Test 077 was a crook: It passed so far because install was expected to fail because of denial from user input, but it failed because of the depercation warning -> I deleted it. There is still 076 which tests non-single...
- Test 116 became more general as curl/wget is not used anymore -> since lfs images also run now we have some move&disabled lfs hooks, therefore the small changes. 
- All other test changes for adapted update triggern

Thanks for quickly reviewing. Tests pass.


gabyx:
Tests all pass, would be good to go :+1:

gabyx:
Rebased, thanks for the pace,and your time
In that I can make some progress. Otherwise its quite cumbersome, you in australia me in europe. hm...

gabyx:
Also eveything works: I think I need to look closer to the update from previous version since we have git now I can directly enforce this easily in a a test

gabyx:
Could you please open a 'dev' branch and add me to the collaborators such that I can push and merge PRs directly into dev and you can coordinate the update to master which is the release. In this way I can develop a version which consists of several pulls you can review too but dont need to be active all the time.

gabyx:
Updates from old versions should be fine. I will add a old version update test in another PR.  could you merge this. Thanks 🙏

rycus86:
> Rebased, thanks for the pace,and your time
In that I can make some progress. Otherwise its quite cumbersome, you in australia me in europe. hm...

Thanks a lot for your persistence! I try to have a look at this in the morning before my day starts, then I van mostly only have a look at night again (around this time) sadly.

> Could you please open a 'dev' branch and add me to the collaborators such that I can push and merge PRs directly into dev and you can coordinate the update to master which is the release. In this way I can develop a version which consists of several pulls you can review too but dont need to be active all the time.

I was thinking about something like this too. We can try it to see how it works. I'm a bit concerned how easy would it be to make review changes though if your branch is 5 PRs ahead of reviews. 🙂 Any suggestions for that case maybe?

base-template.sh
```diff
@@ -793,36 +779,68 @@ show_prompt() {
...
+        )
+
+        # shellcheck disable=SC2181
+        if [ $? -ne 0 ]; then
+            echo "! Fetching updates in  \`$GITHOOKS_CLONE_DIR\` failed with:" >&2
+            echo "$FETCH_OUTPUT" >&2
+            return 1
+        fi
+
+        CURREN_COMMIT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-parse "$GITHOOKS_CLONE_BRANCH")
```
rycus86, line 816: Typo, should `CURRENT_COMMIT`

base-template.sh
```diff
@@ -1000,7 +941,7 @@ read_single_repo_information() {
 #   0 if they were, 1 otherwise
 #####################################################
 is_single_repo() {
-    [ "$IS_SINGLE_REPO" = "yes" ] || return 1
+    [ "$(git config --get --local githooks.single.install)" = "yes" ] || return 1
```
rycus86, line 944: Should this also deal with `true` ?

base-template.sh
```diff
@@ -1011,12 +952,21 @@ is_single_repo() {
...
-        return 0
+    if is_update_available; then
+
+        MESSAGE="$(printf "%s\n%s" \
+            "* There is a new Githooks update available: Forward-merge to commit \"$(echo "$UPDATE_COMMIT" | cut -c1-7)\"" \
+            "Would you like to install it now?")"
+        show_prompt EXECUTE_UPDATE "$MESSAGE" "(Yes, no)" "Y/n" "Yes" "no"
+
+        if [ -z "$EXECUTE_UPDATE" ] || [ "$EXECUTE_UPDATE" = "y" ] || [ "$EXECUTE_UPDATE" = "Y" ]; then
+            return 0
```
rycus86, line 964: So at this point  we have already updated the release checkout.
What's this prompt for? If you say no, we won't revert, is that right?

rycus86: Oh, I see this would be followed by running the new install script from the checkout.
So today this works, when we switch to wrappers and such, we wont.
Can't we do an update check with `git fetch` or something, to see if the remote has a newer commit?

rycus86: And I just saw that this is exactly what you do. 🤦‍♀️
Sorry, misread it.

rycus86: OK, I found the `fetch` but where do we actually pull/merge those changes when we say yes?

cli.sh
```diff
@@ -1562,47 +1598,34 @@ is_git_repo() {
...
+        "$@"
+}
+
+#####################################################
+#  Creates the release clone if needed.
 # Returns:
 #   1 if failed, 0 otherwise
 #####################################################
-update_release_clone() {
+assert_release_clone() {
```
rycus86, line 1622: I'd leave this as `update` rather than `assert`, because it doesn't only assert but can also clone as a side effect.

install.sh
```diff
@@ -1331,43 +1377,110 @@ update_release_clone() {
...
+        echo "Fetching Githooks updates ..."
+        FETCH_OUTPUT=$(
+            execute_git "$GITHOOKS_CLONE_DIR" fetch origin "$GITHOOKS_CLONE_BRANCH" 2>&1
         )
 
         # shellcheck disable=SC2181
         if [ $? -ne 0 ]; then
-            echo "! Pulling updates in \`$CLONE_DIR\` failed with:" >&2
-            echo "$PULL_OUTPUT" >&2
+            echo "! Fetching updates in  \`$GITHOOKS_CLONE_DIR\` failed with:" >&2
```
rycus86, line 1429: Double space
```suggestion
            echo "! Fetching updates in \`$GITHOOKS_CLONE_DIR\` failed with:" >&2
```

install.sh
```diff
@@ -1331,43 +1377,110 @@ update_release_clone() {
...
+            echo "$FETCH_OUTPUT" >&2
             return 1
         fi
-    else
-        clone_release_repository || return 1
+
+        CURREN_COMMIT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-parse "$GITHOOKS_CLONE_BRANCH" 2>/dev/null)
```
rycus86, line 1434: Same typo as above.

rycus86:
What's the reason for doing the update (merge) in the install script? I'd have expected that we update  the release checkout either daily (from the base template) or from the cli, then we execute the install script of the new commit. Would that work? Trying to understand the reasoning for you choice.

gabyx:
https://github.com/rycus86/githooks/pull/106#discussion_r436261222
Ehm I forgot this true false for single.install, but we leave it as it falls out anyway in the future

the reasoning to do the merge ind the install script is that its better to do it there because the earlier we merge a new update it can be problematic since we cannot control how the update is gonna proceed from version before to the version merged, think about some legacy patches we apply and other stuff like deprecation warning, if we updated in the basetemplate already its hard to rollback and so on. I rather have everything at the same place for better controll and reasoning. therefore i strongly recommend of having the update procedure at one place. does that make sense. we also would like later to have proper control from where to where we are updating (sha hash) etc

gabyx:
thx for the review will make the changes later

gabyx:
the assert_release_clone was thought like: assert that we have a release clone. but never update. if it would be called „update“ than IMO its less clear, does it do fetch and merge too? no it doesnt it just clones... in the case its not there hmm

base-template.sh
gabyx: Forgot that, and dont need to change this because we any way gonna remove it.

gabyx:
Ok summary:
We have the following scenarios:

- **Update from cli.sh**
  - git hooks install 
       1. launch install.sh
       2. update the clone 
       3. **dispatch** to the new fetched script
       4. install.sh  -> into this repo (currently setting the single flag -> standalone)
  - git hooks install --global
       1. launch install.sh
       2. update the clone 
       3. **dispatch** to the new fetched script
       4. install.sh  -> global install
- **Update from base-template.sh** daily
       1. check if we have a fetched update availabe ... if yes
       1. launch install.sh
       2. update the clone 
       3. **dispatch** to the new fetched script
       4. install.sh  -> into this repo (currently setting the single flag -> standalone)

gabyx:
I improved the logic in `install.sh:execute_installation` in the last push and integrated the review comments.

cli.sh
gabyx: the side effect of cloning was really meant to be the assert. Its ok to clone, but its not ok to do an update on the clone if there would be... so I leave this

gabyx:
All tests pass -> Good to go :+1:

gabyx:
Next PR would be:
Remove the version number from the 3 files and place it somewhere else in the repo
Less merge conflicts and so during developement ...

gabyx:
FYI: I just tests this and it works:
```shell
git checkout master~50
./install.sh
git hooks install --global
```
-> works

gabyx:
@rycus86  Can you try the following:
```shell
cd githooks
./uninstall.sh
rm -rf ~/.githooks
git config --global githooks.autoupdate.updateCloneUrl https://github.com/gabyx/githooks.git
git config --global githooks.autoupdate.updateCloneBranch feature/update-inside-install
# Install old ....
git checkout master~50
./install.sh
```
Put `echo "OLD CLI"` and `set -x` in `~/.githooks/bin/githooks`
```shell
git hooks install --global
```
It seems that because the new install.sh overwrites the `cli.sh` some f**** up behavior happens 
it executes the cli command `install hook` a second time!! just random coincidence...
I will change the update of the cli, to `rm cli.sh` and then `add a new file` this should fix this weird stuff...

gabyx:
ok good to go.

gabyx:
Rebased onto master, thanks, lets see if the tests still work. Had to adjust test 094 which gave some merge conflicts...

gabyx:
https://github.com/rycus86/githooks/pull/106#issuecomment-640106784 -> Last commit reverted since this is not needed anymore.

gabyx:
Tests fail due to 094, probably made a mistake in the merge and adjustements... will fix it later.
If you have time, take a look :-)

gabyx:
Shall we adjust the indentation for continuation lines in a next PR quickly maybe...

install.sh
```diff
@@ -1344,43 +1395,127 @@ update_release_clone() {
...
-            [ "$BRANCH" != "$GITHOOKS_CLONE_BRANCH" ] ||
-            ! git -C "$CLONE_DIR" diff-index --quiet HEAD; then
-            PULL_ONLY="false"
+            [ "$BRANCH" != "$GITHOOKS_CLONE_BRANCH" ]; then
+
+            CREATE_NEW_CLONE="true"
+
+            # During an autoupdate we would silently erase/reclone
+            # which is not so good, therefore we abort here.
+            if is_autoupdate; then
```
rycus86, line 1410: Will this work for people updating to this version?
Do we already have those variables set to the right branch and URL?

tests/coverage.sh
```diff
@@ -43,13 +43,14 @@ RUN \\
...
-    sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh 
+    sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh
 
 # Commit everything
 RUN echo "Make test gitrepo to clone from ..." && \
     cd /var/lib/githooks && git init && \
     git add . && \
-    git commit -a -m "Initial release"
+    git commit -a -m "Initial release" && \
+    git commit -a --allow-empty -m "Empty to reset to trigger update"
```
rycus86, line 53: Why is this necessary now?

tests/step-116.sh
```diff
@@ -131,14 +126,19 @@ find /tmp -type f -path "*/.git/hooks/*" |
...
+
 cd /tmp/test116.3 &&
-    sed -i 's/^# Version: .*/# Version: 0/' /tmp/test116.3/.git/hooks/post-commit &&
     git config --global githooks.autoupdate.enabled true &&
     git config --global githooks.autoupdate.lastrun $MOCK_LAST_RUN &&
     git commit --allow-empty -m 'Second commit' || exit 1
 
 # Check that all hooks are updated
-find /tmp -type f -path "*/.git/hooks/*" |
+find /tmp -type f -path "*/.git/hooks/*" -and -not -name "*disabled*" |
```
rycus86, line 141: What made this change necessary now?

rycus86:
> Shall we adjust the indentation for continuation lines in a next PR quickly maybe...

Yeah, don't worry about them tok much. Not really sure now how I set them up initially, might have been coincidence/chance with some of those indentations. 🙈

tests/step-116.sh
rycus86: Sorry, saw in the PR description now. 👍

rycus86:
Merged, thanks a lot!
Have a look at the two questions please, but I think we should be fine.

install.sh
gabyx: From really really old scripts we probably not pass the test `if is_autoupdate` because the update doesnt pass the flag to install.sh. But should not matter we then just clone from **this** (default url) repo as a default. For other version already passing the flag to `install.sh` it should work.
(As a sidenote we should maybe think about deprecating update from too old version, to get our head clean of crazy thoughts about what all can happen, I dont know yet a strategy to detect from where we are updating... to deny the update if we are too old and then asking the user to uninstall and reinstall a latest version... etc)

tests/coverage.sh
gabyx: haha, to really easy trigger an update ;-)

gabyx: we reset back one commit :-)

gabyx: and trigger an update in this way which is really easy now... no more fiddeling with versions...

gabyx:
Next PR: Remove Version from the 3 scripts -> putting it into `VERSION` or what woud you like `Version.md` ?

gabyx:
Thanks for merging.

rycus86:
> Next PR: Remove Version from the 3 scripts -> putting it into VERSION or what woud you like Version.md ?

Yeah, `VERSION` sounds good to me. Alternatively, maybe let's think about dropping a hard-coded version again. I'd like a `git hooks version` maybe that generates the same version string as it does today, perhaps some formatting with `git log`, but I can see how these merge conflicts are not helping. So if you think that's something you would rather have a look at, let's try that?

PR #108, opened 2020-06-07
Remove version

gabyx:
Not yet ready.


gabyx:
The tests listed here habe some Version resetting which did not get ported to a git reset unfortunately. Do you understand why these test still worked :)? I think because you only test the autoupdate.lastcheck variable which is not really a test about if the update was executed. i think we rather should concentrate on having a few tests but proper ones. I am not sure about that. What do you think and were the tests intended for something else ?

rycus86:
I think the previous code was only setting that variable after a successful update if I remember correctly, perhaps that's why it worked in the first place. Not sure why the new code made it still pass, have a look at https://coveralls.io/github/rycus86/githooks if you're interested, maybe the coverage sheds a light on which paths are tested since the update mechanism changed.

> i think we rather should concentrate on having a few tests but proper ones. I am not sure about that.

I think we could test more, but didn't want to ask you to do that on top of all the great work you're doing here already. I'm not too concerned for now as we don't get too many bugs here, most of them are found by you pretty quickly if they slip through, so that's OK I think. If it gets hard to make changes because things break without tests, then we should revisit this. On the other hand, I wouldn't start deleting tests just yet, they were covering all of the functionality more or less, which didn't change massively, only their implementation. I might have written some tests badly though that assumed too much about the internals, which is we those might give you trouble now. 🙈

gabyx:
Hm.. sound good to me:
The question is the 
```shell
check_for_updates_if_needed() {
    read_last_update_time
    should_run_update_checks || return
    record_update_time <<<<<<<<<<<<<<<<<<<<HERE
    fetch_latest_updates || return
    should_run_update && execute_update && return
    print_update_disable_info
}
```
I cannot remember that thing changed the order? Currently, this is before any update happened?
Should that be moved after the update really took place? or is `lastrun` only for denoting the last check for update triggern?

gabyx:
Ok, from what I read in `read_last_update_time` its just the update check time...

rycus86:
Yeah, I can't remember exactly now. From what you pasted here, it does look like it was maybe testing the wrong thing. With the new update perhaps comparing commits is better, I think you've done this already in a few tests.

gabyx:
As I found out:
https://unix.stackexchange.com/questions/591618/how-to-supress-the-error-of-file-redirection

`read $VARIABLE </dev/tty` would have been enough since that nasti tty cannot open output is still there :-) So long story short it would have been easier than we discussed...

gabyx:
I removed the VERSION completly since its not needed. We look up the version in the cli.sh and elsewhere for the output with `git rev-parse`. If we would introduce some tags with could use `git describe` which is nice :-).

With that PR we cannot see the version anymore in the installed base-template.sh. 
But hopefully that will drop, maybe we can write the version number to the wrapper script with a post-commit hook, whenever the wrapper is changed... later

gabyx:
Output
image

gabyx:
By the way you should try oh-my-zsh (if you're not already useing it) and https://github.com/denysdovhan/spaceship-prompt
and code highligthing which is really neat, you make less errors :-) in the shell 
https://computingforgeeks.com/how-to-configure-zsh-syntax-highlighting-on-linux-macos/#:~:text=Zsh%20syntax%20highlighting%20enables%20highlighting,particularly%20in%20catching%20syntax%20errors.&text=Syntax%20highlighting%20is%20done%20by%20pluggable%20highlighter%20scripts.

cli.sh
```diff
@@ -2538,12 +2539,13 @@ git hooks version
         return
     fi
 
-    CURRENT_VERSION=$(grep -E "^# Version: .*" "$0" | head -1 | sed -E "s/^# Version: //")
-
+    CURRENT_VERSION=$(execute_git "$GITHOOKS_CLONE_DIR" rev-parse --short HEAD)
```
rycus86, line 2542: Would it be difficult to recreate the version number we had before?
```
VERSION_DATE=$(date +%y%m.%d%H%M)
VERSION_COMMIT=$(git rev-parse --short=6 HEAD)
```

base-template.sh
```diff
@@ -4,6 +4,7 @@
 # It allows you to have a .githooks folder per-project that contains
 # its hooks to execute on various Git triggers.
 #
+# Legacy version number. Not used anymore, but old installs read it.
 # Version: 2006.062037-dd10de
```
rycus86, line 8: Can you please change this something somewhat invalid?
Maybe `Version: 2006.000000-000000` just so that it doesn't look like something we should be trying to interpret.

rycus86:
LGTM, please check the two comments above, then we should be good to merge. 👍

rycus86:
Also wanted to say: thank you very much for these small PRs, much easier to review them! 👌🙂

cli.sh
gabyx: But what would that mean :-) we always get a new version number?
I think it should be git based on tags (semantic versioning)... or then simply the changeset number?

base-template.sh
gabyx: The problem is, old update still read this, and the need to be sure its newer ?
Would you like something like `6666.000000-000000`

gabyx: Think we need somthing like `# Version: 9912.310000-000000`

gabyx: to make some sense

cli.sh
rycus86: Sorry, should have been clearer. I thought we could ask `git log` to output the date of the current HEAD commit.

rycus86: I'll leave it up to you, if this is easy, it might be nice.
Otherwise just resolve the conflict please then I'll merge this PR. 👌

rycus86:
I've added a commit to try and be more consistent with the version number display.
I've also fixed the Arch tests. :partying_face: 
Not sure what's wrong with `ls` there, but `find -type f` worked instead. :man_shrugging:

PR #110, opened 2020-06-11
Bugfix/fail in fetch

gabyx:
- Add a fail message if url/branch is not correct during fetch. This is needed, we already do this in install as well but the fetch can also fail stupidly if we do not check before if everything is correct as we want it. Its not mission critical but gives bad output
- Rename wrong cli.sh command `update-clone-branch` and `update-clone-url` to reflect the 

Thanks for merging.


gabyx:
All tests pass, good to go

PR #111, opened 2020-06-11
Feature: Symlink base-template.sh

gabyx:
Symlink base-template.sh

as a Bonus:
- hooks folder with all generated hooks for simple installing (cloning and setting `core.hooksPath`)


rycus86:
Looking forward to this! Can we please make sure we don't also put symlinks and other changes in this PR?

gabyx:
Jeah, good point, I try to remove the symlinks first. Will make things easier also for me

gabyx:
Lets see what we can do now with the symlinks, that PR we have to discuss thoroughly. Lets see if it works...

gabyx:
So far we get some troubles with the symlinks (?)
```
trace: exec: git-lfs post-checkout /root/.githooks /tmp/test103-clone/.git/hooks/post-checkout 0000000000000000000000000000000000000000 674229b3a048ab86f8486367a7f7a5942974aa5b 1
14:02:57.051695 run-command.c:663       trace: run_command: git-lfs post-checkout /root/.githooks /tmp/test103-clone/.git/hooks/post-checkout 0000000000000000000000000000000000000000 674229b3a048ab86f8486367a7f7a5942974aa5b 1
This should be run through Git's post-checkout hook.  Run `git lfs update` to install it.
```

gabyx:
Ah found it. Forgot to shift arguments

gabyx:
Now tests should all pass.

gabyx:
@rycus86 : The problem with the tests are the following:
If we want to go down the road of having wrappers placed, we have the adjusted the arguments to `base-template.sh` (rather naming this at the end `githooks.sh`)
We should rewrite all tests (might get the PR to overboard, but its the only solution)
from
```
```
HOOK_NAME=pre-commit HOOK_FOLDER=$(pwd)/.git/hooks \
        sh /var/lib/githooks/base-template.sh
```
to 
```
sh /var/lib/githooks/base-template.sh $(pwd)/.git/hooks/pre-commit 
```
or directly calling the 
```
INSTALL_DIR="..."  HOOK_NAME=pre-commit HOOK_FOLDER=$(pwd)/.git/hooks
    sh base-template-symlink.sh
``` 
It gets nasty, but doable...

@rycus86  What do you think?

**Comment by rycus86** [comment]
I'll have a look later today, can you please hold off until then? I need to understand the problem better first. Thanks!

**Comment by gabyx** [comment]
Jeah for sure!, Just look at the tests, alpine works, debain doesnt... etc

**Comment by rycus86** [comment]
A quick Googling says we could just fix the shift rather than rewriting all the tests: https://superuser.com/a/897149

Haven't looked properly yet though, but could be something to look into in the meantime.

File: base-template-symlink.sh [hunk 1]
```diff
@@ -0,0 +1,23 @@
...
+    echo "  Install directory at '${INSTALL_DIR}' is missing." >&2
+    echo "  Please run the Githooks install script again to fix it." >&2
+    exit 1
+elif [ ! -f "$GITHOOKS_TEMPLATE" ]; then
+    echo "! Githooks link to '$GITHOOKS_TEMPLATE' is broken!"
+    echo "  Please run the Githooks install script again to fix it." >&2
+    exit 1
+fi
+
+# shellcheck disable=SC1090
```
rycus86, line 22: Instead of this, could you do the recommendation from https://github.com/koalaman/shellcheck/wiki/SC1090 ?

base-template-symlink.sh
```diff
@@ -0,0 +1,23 @@
...
+sh "$GITHOOKS_TEMPLATE" "$0" "$INSTALL_DIR" "$@"
```
rycus86, line 23: We could have `GITHOOKS_TRIGGER_NAME=$(basename "$0") GITHOOKS_INSTALL_DIR=$IMSTALL_DIR sh "$GITHOOKS_TEMPLATE" "$@"` and then handle the environment variables rather than shifting arguments. What do you think?

tests/step-103.sh
```diff
@@ -83,7 +83,7 @@ fi
 # Clone a new one
 echo "Cloning"
 cd /tmp || exit 1
-git clone /tmp/test103 test103-clone && cd test103-clone || exit 1
+GIT_TRACE=1 git clone /tmp/test103 test103-clone && cd test103-clone || exit 1
```
rycus86, line 86: What does `GIT_TRACE=1` do?

gabyx: Ups left over. From Debugging, it just outputs Traces... quite handy..

base-template-symlink.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+# Base Git hook template from https://github.com/rycus86/githooks
+#
+# It allows you to have a .githooks folder per-project that contains
+# its hooks to execute on various Git triggers.
+#
```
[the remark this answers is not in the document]
gabyx: Ah jeah :-)

.githooks/pre-commit/update-templates
```diff
@@ -0,0 +1,28 @@
...
+
+for HOOK_NAME in $HOOK_NAMES; do
+    if ! echo "$HOOK_NAMES" | grep -q "$HOOK_NAME"; then
+        echo "! You changed the managed hooks types" >&2
+        echo "  you need to adjust them in \`$0\`" >&2
+        exit 1
+    fi
+
+    cp base-template-symlink.sh "hooks/$HOOK_NAME"
+    if ! sed -i "s|INSTALL_DIR=.*|# Auto-generated: do not edit!\nINSTALL_DIR=\$(cd \"\$(dirname \"\$0\")/../\" \&\& pwd)|" "hooks/$HOOK_NAME" ||
```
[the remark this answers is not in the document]
gabyx: Ehm, did you note that the hooks in `hooks/` are only here for a quick install (as @matthijskooijman suggested) I though of having these checked in for that sole purpose...

We actuall still install the wrappers `base-template-symlink.sh` into the repos `.git/hooks` dir.

Do you mean when we place the wrapper `base-template-symlink.sh` in the repos, we directy replace the INSTALL_DIR? Does that make sense, I rather have looked it up in the config... ?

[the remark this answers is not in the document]
gabyx: See comment blow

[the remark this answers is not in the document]
gabyx: Ok, might be :-)

base-template-symlink.sh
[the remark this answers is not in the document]
gabyx: Hm.. since we `execute_hook ... || return 1` in  `base-template.sh` everywhere, I assume we can do ` || return 1` here right?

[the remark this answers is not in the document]
gabyx: I think we inherit the exit code ...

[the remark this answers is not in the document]
gabyx: Jeah, but the point was to try (at least) not to rely anymore on env variables, its kind of nasty, but your approach would work! What do you think?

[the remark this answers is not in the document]
gabyx: Hm... I thought of with shifting
```shell
HOOK_NAME=$(basename "$0")
HOOK_FOLDER=$(dirname "$0")
sh "$GITHOOKS_TEMPLATE" "$HOOK_NAME" "$HOOK_FOLDER" "$INSTALL_DIR" "$@"
```
and I think no tests work, because how can we know how much to shift leaving us in rewriting all `base-template.sh` calls in the tests (39 in total)
or 
```shell
HOOK_NAME=$(basename "$0")
HOOK_FOLDER=$(dirname "$0")
GITHOOKS_HOOK_NAME="$HOOK_NAME" GITHOOKS_INSTALL_DIR="$INSTALL_DIR" GITHOOKS_HOOK_FOLDER="$HOOK_FOLDER" sh "$GITHOOKS_TEMPLATE" "$@"
```
with no shifting and all tests work...

Hm...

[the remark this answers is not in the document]
gabyx: I'd rather would have proper arguments instead of env variables. Any good hooks manager does it like that I think -> also `git lfs post-merge ...`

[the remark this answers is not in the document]
gabyx: or we replace all calls to `base-template.sh` in the tests with `base-template-wrapper.sh` and do some trickery replacement in this file in the dockerfile... that might work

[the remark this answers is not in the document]
gabyx: hm... an env. approach might be the least trouble some...

.githooks/pre-commit/update-templates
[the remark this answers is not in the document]
rycus86: Will that even work? Won't we need the install dir settings to exist when running hooks? Like for checking for updates, etc. Maybe we should just do the symlink change here, but ignore the pre-generated hooks in this first round, so we can see how those work properly in a next PR? What do you think?

base-template-symlink.sh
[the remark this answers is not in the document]
rycus86: I'd vote for environment variables, I don't think they are too bad. Do you have anything in particular that you know about that makes them "nasty"?
I think we get some env variables from Git itself too, don't we?
In any case, I think using them might be the least amount of change for now, then we can come back to arguments vs. env variables later?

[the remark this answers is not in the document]
rycus86: Can you make sure? With a test perhaps?

base-template-symlink.sh
```diff
@@ -0,0 +1,24 @@
...
+    echo "  Please run the Githooks install script again to fix it." >&2
+    exit 1
+elif [ ! -f "$GITHOOKS_SCRIPT" ]; then
+    echo "! Githooks link to '$GITHOOKS_SCRIPT' is broken!"
+    echo "  Please run the Githooks install script again to fix it." >&2
+    exit 1
+fi
+
+# shellcheck disable=SC1090
+sh "$GITHOOKS_SCRIPT" "$INSTALL_DIR" "$0" "$@"
```
matthijskooijman: This assumes that the script uses sh, which would prevent ever switching to .e.g. `bash`. Wouldn't it also work to omit `sh` here (assuming the script is executable in git and has a `#!` line)?

base-template.sh
```diff
@@ -15,7 +15,8 @@
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
-    set_main_variables "$@"
+    set_main_variables "$@" && shift && shift
```
matthijskooijman, line 18: This looks a bit unclear to me. Would it be better to shift at the place where `$1` and `$2` are interpreted? Or set them where the `shift` is?

Also, a comment about the shift might be helpful (e.g. shift off extra arguments added by symlink, leaving only the original git-provided arguments in `$@`).

base-template.sh
```diff
@@ -61,17 +62,22 @@ are_githooks_disabled() {
 # Returns: None
 #####################################################
 load_install_dir() {
-    INSTALL_DIR=$(git config --global --get githooks.installDir)
+
+    # INSTALL_DIR might be set from base-template-symlink.sh
```
matthijskooijman, line 66:
```suggestion
    # INSTALL_DIR might be passed from base-template-symlink.sh
```
I was a bit confused by the word "set", thinking it would actually set the variable directly (through environment or so).

.githooks/pre-commit/update-templates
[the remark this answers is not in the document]
matthijskooijman: I'm still hoping githooks can be made to run without any installation other than setting `core.hooksPath`, but then indeed `INSTALL_DIR` must be derived from the script location rather than git config, so this replace seems reasonable to me (since the pregenerated version needs the derive-from-$0 while the installed version needs get-from-git-config).

But indeed, this is rather orthogonal to the change in this PR (unless this change would need changes in the symlink and more importantly the symlink-to-base-template calling interface, since that cannot be easily changed once merged, I think).

base-template.sh
```diff
@@ -89,8 +95,11 @@ load_install_dir() {
 # Returns: None
 #####################################################
 set_main_variables() {
-    HOOK_NAME=$(basename "$0")
-    HOOK_FOLDER=$(dirname "$0")
+
+    HOOK_NAME=$(basename "$1")
+    HOOK_FOLDER=$(dirname "$1")
+    INSTALL_DIR="$2"
```
matthijskooijman, line 101: Is it really needed to pass INSTALL_DIR here? It slightly complicates the symlink-to-base-template interface and the symlink script and might introduce a chance of mismatch? Did you consider the "derive from $0-trick` I suggested [originally](https://github.com/rycus86/githooks/pull/92#discussion_r424998436)?

base-template-symlink.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/sh
+# Base Git hook template from https://github.com/rycus86/githooks
+#
+# It allows you to have a .githooks folder per-project that contains
+# its hooks to execute on various Git triggers.
+#
+# Version: 9912.310000-000000
+
+INSTALL_DIR=$(git config --global githooks.installDir)
+GITHOOKS_SCRIPT="$INSTALL_DIR/release/base-template.sh"
```
matthijskooijman: Would it make sense to just store the full path to `base-template.sh` in a git config variable instead of deriving this from the `installDir`? That would prevent the need for hardcoding this filename and path here (i.e. the path within the git dir) and simplify the symlink script (and simpler means less likely to need changes in the future, which is good for a script that has duplicates spread around the filesystem).

base-template-symlink.sh
rycus86: Good point, the base template checks the executable flag on the targets and only uses `sh` if it's not an executable.

base-template.sh
rycus86: Do you know if `$0` is POSIX and expected to work under any shells?

base-template-symlink.sh
rycus86: The script location is a derived value, for now at least, I don't it's necessary to introduce this duplication in the config yet. I'd do this when there's a need for it from the implementation.

rycus86:
> Also, I suggest not using the name "symlink" for this new file. An actual "symlink" is a different thing on Unix, which could have been used instead of the wrapper/stub that you're using now, but for portability reasons (i.e. Windows has no perfect equivalent of symlinks) we're not doing that. Calling this file `base-template-symlink.sh` would likely confuse people into thinking symlinks were somehow used.

I agree, symlink might be a bit misleading. 

> Also, I would suggest renaming `base-template.sh`, since now it stays in the git repository and is not copied anywhere, it's not really a template any more.

I'm not too fussed about this, it is still a thing other things refer to. If you find a name that describes it really well, let's change it, otherwise I'd leave it. 🤷‍♂️

> How about `hook-runner.sh` instead of `base-template.sh` (since that's what it does, it runs hooks when called)?

It does run hooks, besides other things. 🙂 Not really happy with that name, but I'm also not sure I have anything better at the moment. Keep the ideas coming! 👍

> Then for the "symlink", I was thinking of something like `hook-stub.sh`, `hook-shim.sh` or `hook-wrapper.sh`, but looking at the proper meaning of stub, shim and wrapper, none of them seem a perfect fit. Maybe the symlink could be `hook-template.sh`, since this is now the script that's copied into the different places (even though it's copied verbatim, not really changing anything like you would expect with a template, other than the name of the file maybe).

Also consider `dispatch`, though I think `wrapper` could also work as @matthijskooijman suggested.

base-template.sh
matthijskooijman: https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html#tag_02_05_02

> 0    (Zero.) Expands to the name of the shell or shell script. See sh for a detailed description of how this name is derived.

Then the sh page has more details:

https://pubs.opengroup.org/onlinepubs/009695399/utilities/sh.html

Here, it's relevant how `sh` was invoked. WHen using `#!/bin/sh` AFAIU the kernel just appends the script filename that contains the `#!` to `/bin/sh`, so it invokes this flavor of the `sh` commandline:

    sh [-abCefhimnuvx][-o option][+abCefhimnuvx][+o option] [command_file [argument...]]

So then, this rule triggers:

> Special parameter 0 (see Special Parameters) shall be set to the value of command_file. If sh is called using a synopsis form that omits command_file, special parameter 0 shall be set to the value of the first argument passed to sh from its parent (for example, argv[0] for a C program), which is normally a pathname used to execute the sh utility.

IOW, `$0` is always set to the value of the script for cases we care about (i.e. when using the `#!`, or when just running `sh /path/to/base-template.sh`. There is a way to let `$0` point to something else (by specifying it explicitly in the ` sh -c[-abCefhimnuvx][-o option][+abCefhimnuvx][+o option]command_string [command_name [argument...]]` form), but that is not applicable here.

rycus86: Thanks for checking!

base-template-symlink.sh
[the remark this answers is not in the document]
matthijskooijman: What I don't like about env vars is that they have a global shared namespace and it is somewhat easy to accidentally read them from somewhere else, especially when you're migrating interfaces (e.g. if you add a new env var later, then any existing "symlink" scripts will not set it, but if the env var was already set elsewhere, you'd *think* the symlink script has set it even when it did not.

However, with commandline arguments, change might also be non-trivial, since if you add an argument, you'd have to somehow know that you did.

> In any case, I think using them might be the least amount of change for now, then we can come back to arguments vs. env variables later?

One thing to keep into account when changing is that you might need to support existing "symlink" scripts that still use the old interface. Or are all symlink scripts already updated on update? The idea is to not do that, right?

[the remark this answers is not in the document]
matthijskooijman: You should probably do `exec sh ...` (or `exec "$GITHOOKS_TEMPLATE" ...`) so the template process replaces the current one, instead of forking (saves one fork). In either case, the exit code is inherited, AFAIK.

base-template-symlink.sh
matthijskooijman: But the current approach has a lot more duplication, each installed version of the "symlink" script will now have the `release/base-template.sh` path hardcoded, in a place that is not intended to be updated normally (right?).

To make my suggestion more specific, I think it could be as simple as this:

```
RUNNER=$(git config --global githooks.runner)

if [ ! -x "$RUNNER" ]; then
    echo "! Githooks runner ('$RUNNER' from git config 'githooks.runner') is missing or broken"
    echo "  Please run the Githooks install script again to fix it." >&2
    exit 1
fi

# shellcheck disable=SC1090
exec "$GITHOOKS_SCRIPT" "$0" "$@"
```

I also removed the `sh` and added `exec` as suggested elsewhere, and removed the `$INSTALL_DIR` (assuming base-template will derive that from `$0`). The shellcheck thing still needs to be fixed.

You could simplify even more and just have a single line script:

```
exec "$(git config --global githooks.runner)" "$0" "$@"
```

Which would rely on the shell to generate the error message for a missing or broken runner, but that might be a bit less friendly.

gabyx:
All good comment, just let you know, I want have lots of time in the next 4 Weeks, unfortunately. If anybody wants to make further progress, please go on!. I had some troubles with the tests which made me stumble upon other stuff. I will read your comments later!
Thanks for the review.

rycus86:
> All good comment, just let you know, I want have lots of time in the next 4 Weeks, unfortunately. If anybody wants to make further progress, please go on!. I had some troubles with the tests which made me stumble upon other stuff. I will read your comments later!

No worries @gabyx , thanks a lot for your hard work on these PRs! I'll see of I can have a look at them while you're away. 👍

base-template-symlink.sh
gabyx: Very good, jeah I think, this is nice

gabyx: I wouldnt use `exec`for the first tries, because had troubles with the tests. I would leave that to the end

gabyx: Ive we move to a `githooks.sh` (runner) and a `githooks-dispatch.sh` the above approach would be good I think. Keep it simple stupid.

gabyx:
Some update: I finally continued a bit. Had some changes left in the worktree, Figured out how to do it now.
All tests pass now.

See last commit, for review changes.
- I added `githooks.runner` which is a good point. Makes things easier.
- The manual execution now all execute through the base-template-wrapper.sh (later to be renamed!) to test the whole chain.
- Though, I dont like the direct execution tests, since this makes tests much harder to correct/predict since we use sed replacements and change the code. So in the future it would be somehow better to rewrite them to just use a proper install like the other tests (would there be a problem with that)

@rycus86 Could you have look at this. This needs some serious review I think. I will also test whats gonna happen (:bomb: ) when we update to this code and such...

rycus86:
Thanks a lot for picking this up again @gabyx !
I'll try to find some time later to review your latest changes.

base-template-symlink.sh
gabyx: Applied in latest commit

base-template-symlink.sh
gabyx: This is open, to implement. Will probably break tests again. Keep it open to have a look at it.

.githooks/pre-commit/update-templates
```diff
@@ -0,0 +1,28 @@
+#!/bin/sh
+
+if ! echo "$STAGED_FILES" | grep -q "base-template-symlink|update-templates"; then
+    # stop here if there are no changes for this commit
```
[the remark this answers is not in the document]
gabyx: Jeah, ok, lets take this out.

.githooks/pre-commit/update-templates
```diff
@@ -0,0 +1,28 @@
...
+    exit 0
+fi
+
+# Patch the hooks
+HOOK_NAMES=$(sed -e '1,/MANAGED_HOOK_NAMES="/d;/"/,$d' install.sh)
```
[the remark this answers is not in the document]
gabyx: I think so. its abit dark magic yeah. 
It deletes all lines except the ones we want -> the hooks
I added en exit 1 if it fails...

gabyx:
Tests now all pass, had again to adjust the docker copy command

base-template.sh
```diff
@@ -89,8 +91,10 @@ load_install_dir() {
 # Returns: None
 #####################################################
 set_main_variables() {
-    HOOK_NAME=$(basename "$0")
-    HOOK_FOLDER=$(dirname "$0")
+
+    HOOK_NAME="$GITHOOKS_HOOK_NAME"
```
matthijskooijman, line 95: This could really use a comment stating these are set by the wrapper script, IMHO.

base-template-wrapper.sh
```diff
@@ -0,0 +1,18 @@
...
+
+if [ ! -x "$GITHOOKS_RUNNER" ]; then
+    echo "! Githooks runner points to non existing location:" >&2
+    echo "   \`$GITHOOKS_RUNNER\`" >&2
+    echo "  or it is not executable!" >&2
+    echo " Please run the Githooks install script again to fix it." >&2
+    exit 1
+fi
+GITHOOKS_HOOK_NAME=$(basename "$0") \
+GITHOOKS_HOOK_FOLDER=$(dirname "$0") \
```
matthijskooijman, line 17: I'd consider just passing `$0` here and letting the runner script separate it into hook name and hook folder, to make this script even simpler and make it easier to change things later.

gabyx: yeah totally

gabyx: Lets see, it took some iteration till I arrived here, with tests now passing I can try this:-)

gabyx: Also I use exec now.

base-template-symlink.sh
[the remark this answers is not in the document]
gabyx: Yeah symlink scripts should be as simple as we possible can! With my last commit, I tried that, also the input from @matthijskooijman is good I think. I could also get rid easly of the env variables. As all global variables (and sh itself too because we cannot use `local` variables) are a realy Pain in the Ass... for the reason I wont elaborate here tl;dr...
So rather use function, be stateless as far as we can be (ok, its shell script ;-) lool). Also we had a weird bug because some GIT_DIR env variable was leaking from git and so on inside our githook, there is still a note about this... Thats the sad other story about these kind of stupid variables also they are handy I see :-)

base-template-symlink.sh
gabyx: Done!

base-template-wrapper.sh
gabyx: Done!

.githooks/pre-commit/update-templates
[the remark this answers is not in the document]
gabyx: Changed to `HOOK_NAMES=$(sed -e '1,/MANAGED_HOOK_NAMES="/d;/"/,$d' install.sh) `

.githooks/pre-commit/update-templates
```diff
@@ -0,0 +1,28 @@
...
+
+for HOOK_NAME in $HOOK_NAMES; do
+    if ! echo "$HOOK_NAMES" | grep -q "$HOOK_NAME"; then
```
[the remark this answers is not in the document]
gabyx: Was a mistake

gabyx:
Check the new pregenerated hooks, since we now rely only on GITHOOKS_RUNNER, they seem ok.

- I am only a bit worried about `$0` is this alawys an absolut path or also relative? We had a discussion on this I think @matthijskooijman ?

gabyx (commented):
Cancel this review...

base-template.sh
```diff
@@ -15,7 +15,8 @@
 #   0 when successfully finished, 1 otherwise
 #####################################################
 process_git_hook() {
-    set_main_variables
+    set_main_variables "$@" || return 1
+    shift 1
```
matthijskooijman, line 19: Maybe just pass `$1` here rather than `$@`? And shift earlier? So something like:

```suggestion
    ORIGINAL_COMMAND=$1
    shift 1
    set_main_variables "$ORIGINAL_COMMAND" || return 1
```

Then the shifting is in the same place as where `$1` gets meaning (not where it is interpreted, that is `set_main_variables`, but that's ok).

tests/exec-tests.sh
```diff
@@ -33,8 +33,8 @@ RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
     # Change the base template so we can pass in the hook name and accept flags
     sed -i -E 's|echo.*Hook not run inside a git repository.*|CURRENT_GIT_DIR=".git"|' /var/lib/githooks/base-template.sh && \\
     sed -i -E 's|GITHOOKS_RUNNER=(.*)|GITHOOKS_RUNNER=\1; GITHOOKS_RUNNER="\${GITHOOKS_RUNNER:-/var/lib/githooks/base-template.sh}"|' /var/lib/githooks/base-template-wrapper.sh && \\
-    sed -i -E 's|GITHOOKS_HOOK_FOLDER=(.*)\\\|GITHOOKS_HOOK_FOLDER=\1 GITHOOKS_HOOK_FOLDER="\${HOOK_FOLDER:-\$GITHOOKS_HOOK_FOLDER}" \\\|' /var/lib/githooks/base-template-wrapper.sh && \\
-    sed -i -E 's|GITHOOKS_HOOK_NAME=(.*)\\\|GITHOOKS_HOOK_NAME=\1 GITHOOKS_HOOK_NAME="\${HOOK_NAME:-\$GITHOOKS_HOOK_NAME}" \\\|' /var/lib/githooks/base-template-wrapper.sh && \\
+    sed -i -E 's|HOOK_FOLDER=(.*)|HOOK_FOLDER="\${HOOK_FOLDER:-\1}"|' /var/lib/githooks/base-template.sh && \\
+    sed -i -E 's|HOOK_NAME=(.*)|HOOK_NAME="\${HOOK_NAME:-\1}"|' /var/lib/githooks/base-template.sh && \\
```
matthijskooijman, line 37: Why modify these `HOOK_FOLDER` and `HOOK_NAME` lines at all? Why not just modfiy the tests steps to pass this stuff in the normal way? e.g.

```diff
- HOOK_NAME=pre-commit HOOK_FOLDER=$(pwd)/.git/hooks \
-        sh /var/lib/githooks/base-template.sh
+        sh /var/lib/githooks/base-template.sh "$(pwd)/.git/hooks/pre-commit"
```

tests/exec-tests.sh
```diff
@@ -33,8 +33,8 @@ RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
     # Change the base template so we can pass in the hook name and accept flags
     sed -i -E 's|echo.*Hook not run inside a git repository.*|CURRENT_GIT_DIR=".git"|' /var/lib/githooks/base-template.sh && \\
     sed -i -E 's|GITHOOKS_RUNNER=(.*)|GITHOOKS_RUNNER=\1; GITHOOKS_RUNNER="\${GITHOOKS_RUNNER:-/var/lib/githooks/base-template.sh}"|' /var/lib/githooks/base-template-wrapper.sh && \\
```
matthijskooijman, line 35: Maybe unrelated, but `GITHOOKS_RUNNER` does not seem to appear in `base-template.sh`, so this looks like a noop?

tests/exec-tests.sh
```diff
@@ -33,8 +33,8 @@ RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
     # Change the base template so we can pass in the hook name and accept flags
     sed -i -E 's|echo.*Hook not run inside a git repository.*|CURRENT_GIT_DIR=".git"|' /var/lib/githooks/base-template.sh && \\
```
matthijskooijman, line 34: Maybe unrelated, but why is this needed? All tests steps I looked at create a proper git repo to run inside?

tests/exec-tests.sh
```diff
@@ -33,8 +33,8 @@ RUN sed -i 's|</dev/tty||g' /var/lib/githooks/install.sh && \\
...
-    sed -i -E 's|GITHOOKS_HOOK_FOLDER=(.*)\\\|GITHOOKS_HOOK_FOLDER=\1 GITHOOKS_HOOK_FOLDER="\${HOOK_FOLDER:-\$GITHOOKS_HOOK_FOLDER}" \\\|' /var/lib/githooks/base-template-wrapper.sh && \\
-    sed -i -E 's|GITHOOKS_HOOK_NAME=(.*)\\\|GITHOOKS_HOOK_NAME=\1 GITHOOKS_HOOK_NAME="\${HOOK_NAME:-\$GITHOOKS_HOOK_NAME}" \\\|' /var/lib/githooks/base-template-wrapper.sh && \\
+    sed -i -E 's|HOOK_FOLDER=(.*)|HOOK_FOLDER="\${HOOK_FOLDER:-\1}"|' /var/lib/githooks/base-template.sh && \\
+    sed -i -E 's|HOOK_NAME=(.*)|HOOK_NAME="\${HOOK_NAME:-\1}"|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|ACCEPT_CHANGES=|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|'  /var/lib/githooks/base-template.sh && \\
     sed -i 's%read -r "\$VARIABLE"%eval "\$VARIABLE=\\\\\$\$(eval echo "\\\\\$VARIABLE")" # disabled for tests: read -r "\$VARIABLE"%' /var/lib/githooks/base-template.sh && \\
     sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh
```
matthijskooijman, line 40: Maybe unrelated, but why change this url? It seems that base-template.sh and cli.sh take this url from the git config and repo remote settings, which you could just set, and install.sh already uses any value supplied through an env var or commandline option, which could just be passed, without changing the script here?

matthijskooijman (commented):
Nice! Wrapper looks nice and compact now, good to see exec now works. Are the exit -> return changes you made to make the exec work? Or is that unrelated?

I left some more suggestions and remarks inline.

matthijskooijman:
> I am only a bit worried about $0 is this alawys an absolut path or also relative? We had a discussion on this I think @matthijskooijman ?

It depends on how the script is executed. In general, it should either be absolute, or relative to the current directory. I wouldn't be suprised if git makes sure it is absolute always, but if you want to ensure it is absolute, a common approach is something like:

 HOOK_FOLDER="$(cd "$(dirname "$1")" && pwd)"

So change to the folder you're interested in and use `pwd` to print the full path, all in a subshell so the current shell's working dir does not change.

tests/exec-tests.sh
gabyx: Correct, but this is for a later PR, so far we just modify these as before, to *directly* execute the template...
Its lots of regex replace...

tests/exec-tests.sh
gabyx: What url do you mean `/var/lib/githooks/base-template-wrapper.sh ` to `/var/lib/githooks/base-template.sh`?
Its just to inject the proper values for like 30 tests... -> Would be good jup, to adjust these tests too maybe at the end!

tests/exec-tests.sh
gabyx: True, @rycus86 ?: Try to remove it and we will see, Good that we have lots of tests...

tests/exec-tests.sh
gabyx: Look closely its base-template-wrapper.sh :)

tests/exec-tests.sh
matthijskooijman: Hm, wait. It seems that (after this PR) most of the testcases actually seem to call the *wrapper*, rather than the template directly: https://github.com/rycus86/githooks/pull/111/files#diff-07d8a301a5f5b558884d60a40b4c9040

Why not revert that change and apply my above change instead? Just keep calling the script base-template directly, but pass it the right `$0` value? Not sure if that's for another PR, since you're touching this stuff already anyway, and updating the tests to the new way of passing `$0` introduced with this PR seems appropriate to me?

tests/exec-tests.sh
matthijskooijman: I'm talking about this line:

         sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh 

Apparently that fixates `GITHOOKS_CLONE_URL` to `/var/lib/githooks`. I'm not entirely sure *why* this is needed, but I would think you can just do e.g. `install.sh --clone-url /var/lib/githooks` when running install.sh, and do `git config --global githooks.cloneUrl /var/lib/githooks` (and maybe `git remote add origin /var/lib/githooks` after each `git init`) to just set this value normally, so you do not need this `sed` line to fixate the url.

However, this *is* really unrelated to this PR (I just noticed it here), so probably best to leave for now.

tests/exec-tests.sh
matthijskooijman: Ah, indeed :-)

I was going to suggest to just set the global config, which of course works when running in docker / CI, but when running tests separately, you don't want to mess with a user's git config, so that won't really work.

But, this makes me wonder: Can't we just remove `--global` when reading config values, e.g. here: https://github.com/rycus86/githooks/pull/111/files#diff-13c7c6fb70b550a166546e1b5f5269a8R7

That would read the value from the current repo when it is set, or the global otherwise. That would allow setting the config value just for the current test repository, without needing this sed line. As an added bonus, it would also allow users to install multiple versions of githooks side-by-side and configure which one to use on a repo-by-repo basis by using non-global config values (not sure if that would be useful, but seems neat if it would be possible).

tests/exec-tests.sh
matthijskooijman: Ah, setting global git configs is probably not ideal, but like I mentioned in https://github.com/rycus86/githooks/pull/111#discussion_r472838040, reading `githooks.cloneUrl` can probably also happen without `--global`?

tests/exec-tests.sh
gabyx: Jeah, should be appropriate, but lots of work.
I decided to call the wrapper, because it tests the whole chain, and I kind of dont like this "direct" executing of the base-template, and tought, -> call the chain which is also called in the setup nothing else, thats why...

tests/exec-tests.sh
gabyx: Thats true, strange I though there is something in the `exec-steps.sh`

tests/exec-tests.sh
gabyx: Ok, thats right, we can remove --global, since this is better behavior I think!

tests/exec-tests.sh
matthijskooijman: I guess that makes sense. But if you really want to test the entire chain, shouldn't you just install the wrapper into the `.git/hooks` directory and just call that?

Alternatively, it seems you can also call `sh` in a way where it executes one script, but sets `$0` to something else. It's a bit finnicky, but can work like this:
```
matthijs@grubby:~$ cat test.sh 
#!/bin/sh
echo $0
echo $1
matthijs@grubby:~$ sh -c '. ./test.sh' /tmp/.git/hooks/pre-commit some_argument
/tmp/.git/hooks/pre-commit
some_argument
```

So the `-c` argument is a commandline to execute, the rest of the arguments is `$0`, `$1`, etc. For `-c`, I used `.` to source the script, rather than execute it, since the latter would spawn a new shell that would change `$0` to `./test.sh` again.

tests/exec-tests.sh
gabyx: No this line is prefectly fine, I checked now. It makes sense. Its only replacing the **default url** incase none is set.
So it tests more paths the initial guessed in this dicussion.

base-template.sh
gabyx: Done.

tests/exec-tests.sh
gabyx: Done.

tests/exec-tests.sh
gabyx: Seems to ok to removed according to tests... Les replacements -> better.

gabyx:
Incoporated input from @matthijskooijman.

tests/exec-tests.sh
matthijskooijman: Ah, excellent point, thanks for checking!

tests/exec-tests.sh
matthijskooijman: Ok, so now the `--global` is gone, you could also remove this particular `sed` line (and then set the `githooks.runner` variable in each repository created). That's a bit of an invasive change, modifying all test steps AFAICS, so maybe that's better for later.

matthijskooijman:
Looks good, thanks!

tests/exec-tests.sh
gabyx: Yeah, we leave that for another PR. We need also tests, maybe this goes hand in hand with your clone-set-and-use workflow...

gabyx:
@rycus86 Any progress ;-)?

rycus86:
Sorry, I'm super behind on this. 😔
Apologies for the delay, I'll try to get around to reviewing and merging this PR this week, I see that it already has green builds and had a few reviews from @matthijskooijman
Thanks both!

hooks/applypatch-msg
```diff
@@ -0,0 +1,19 @@
+#!/bin/sh
+# Base Git hook template from https://github.com/rycus86/githooks
+#
+# It allows you to have a .githooks folder per-project that contains
+# its hooks to execute on various Git triggers.
+
+# Read the runner script from the local/global or system config
+# Auto-generated: do not edit!
```
rycus86, line 8: /nit: `Read the runner script from the local/global or system config` is not actually happening here which is a bit confusing, but let's maybe look into it as another change.

hooks/applypatch-msg
```diff
@@ -0,0 +1,19 @@
...
+GITHOOKS_RUNNER="$(cd "$(dirname "$0")/../" && pwd)/base-template.sh"
```
rycus86, line 9: What is the actual file this points at?
Is it ./base-template.sh in the Githooks root?
Should it point to the one in ./release ?

rycus86:
Overall LGTM. :+1: 
Just left a couple of comments if you could look at @gabyx

hooks/applypatch-msg
gabyx: it should point too
`base-template.sh` in the githooks root directory (might be a git repo but might not, user may have downloaded a zip version)...
Be aware these hooks are just here to be autogenerated, we do not use them yet, nor do we have tests with which we could tests a simple "clone-and-use" workflow.

hooks/applypatch-msg
gabyx: Will change that, later.

hooks/applypatch-msg
rycus86: OK, if we don't use them then I'm OK to ignore it for now, we'll sort out where these need to point to later.

rycus86:
Awesome stuff, thanks a lot @gabyx ! :tada:

hooks/applypatch-msg
gabyx: Ah you merged already, lool, ok. I will make another knitpik PR for that one... :-)

rycus86: Yeah, since they are not used yet, we can deal with it when we do use them.

PR #112, opened 2020-06-14
Remove single install

gabyx:
Thats the changeset without any symlinks.
Tests seem to work so far. 

We need to check if we made all path from older update clear to fail. I hope we did. Here nothing really much changed. 

I refactored also in this PR 
- moved `githooks.autoupdate.registered` -> `githooks.registered` and moved the file `$INSTALL_DIR/autoupdate/registered` to `$INSTALL_DIR/registered` (maybe other name?)
`legacy_transformtion_end` deletes the whole `$INSTALL_DIR/autoupdate/` folder because that was the only need for it...

Note: its good to do this right here, because the registered list only contained `non-single` installs. And I think the feature to have now **all** repos registered isnt that bad since, we could potentially act on this list in future updates. I would this feature still in. might be handy ...


gabyx:
Shall I also move the autoupdate changes out of this PR. Its a bit more, work, hm... Are you able to look over all this renaming, would be greate, otherwise I try to remove it and add it back in in another PR...? let me know

install.sh
```diff
@@ -298,31 +317,6 @@ parse_command_line_arguments() {
         unset DO_UPDATE_ONLY
         INTERNAL_AUTOUPDATE="true"
     fi
-
-    # Using core.hooksPath implies it applies to all repo's
```
gabyx, line 302: This is no more needed -> `--single` is just a one repo install from now on!

install.sh
```diff
@@ -1192,29 +1171,29 @@ install_hooks_into_repo() {
     else
         # Getting the working tree (no external .git directories)
         # see https://stackoverflow.com/a/38852055/293195
-        TARGET_ROOT=$(git -C "${TARGET}" rev-parse --show-toplevel 2>/dev/null)
+        TARGET_ROOT=$(git -C "$TARGET" rev-parse --show-toplevel 2>/dev/null)
```
gabyx, line 1174: Removed some uneeded braces....

install.sh
```diff
@@ -1234,15 +1213,13 @@ install_hooks_into_repo() {
...
-    [ "$INSTALLED" = "true" ] &&
-        ! is_single_repo_install &&
-        register_repo "$TARGET"
-
     if [ "$INSTALLED" = "true" ]; then
         if is_dry_run; then
             echo "[Dry run] Hooks would have been installed into $TARGET"
         else
+            register_repo "$TARGET"
+            legacy_transformations_repo_end "$TARGET"
```
gabyx, line 1221: Some transformation after we installed...

tests/step-108.sh
```diff
@@ -5,7 +5,7 @@
 TEST_PREFIX_DIR="/tmp/githooks"
 sh /var/lib/githooks/install.sh --prefix "$TEST_PREFIX_DIR" || exit 1
 
-if [ ! -d "$TEST_PREFIX_DIR/.githooks" ]; then
+if [ ! -d "$TEST_PREFIX_DIR/.githooks" ] || [ ! -d "$TEST_PREFIX_DIR/.githooks/release" ]; then
```
gabyx, line 8: Not a fix, just to be a bit more precise...

rycus86:
There are some test failures here, but otherwise the change looks OK, you can leave the config move in this PR if you want.

One question I have is whether we need the legacy transformations in the uninstall script as well? Don't we fetch that with curl? If yes, we'd potentially get a newer version than what you have locally, right?

install.sh
rycus86: /nit: I usually err on the side of having unnecessary braces rather than omitting braces, to avoid problems with missing braces on a refactor. I'm not too fussed about this one in particular here, so I'm fine either way, just wanted to tell you for future changes. :)

gabyx:
nono :-) we never use curl anymore, everything goes over the release clone :-)
the legacy stuff should be in the uninstall script as well to prevent dangling old stuff...

gabyx:
I look at the tests later thanks!

rycus86:
> nono :-) we never use curl anymore, everything goes over the release clone :-)

Hm, https://github.com/rycus86/githooks/blob/master/README.md#uninstalling uses vurl to fetch the uninstall script, does that internally update the release clone now to just delete it?

gabyx:
> > nono :-) we never use curl anymore, everything goes over the release clone :-)
> 
> Hm, https://github.com/rycus86/githooks/blob/master/README.md#uninstalling uses curl to fetch the uninstall script, does that internally update the release clone now to just delete it?

Ah hm.. we might have some inconsistency here, you are right.
You should really call `git hooks uninstall` and not use the curl method.
But we can also check if there is a release clone, if true we dispatch to the install script there... other wise we just continue with the script...?

rycus86:
> > > nono :-) we never use curl anymore, everything goes over the release clone :-)
> > 
> > 
> > Hm, https://github.com/rycus86/githooks/blob/master/README.md#uninstalling uses curl to fetch the uninstall script, does that internally update the release clone now to just delete it?
> 
> Ah hm.. we might have some inconsistency here, you are right.
> One should really call `git hooks uninstall` and not use the curl method preferably?
> But we can also check in `uninstall.sh` if there is a release clone, if true we dispatch to the install script there... other wise we just continue with the script...?

Yeah, I'm not too bothered *how* it happens tbh as long as it works for old and new installations I guess on a best effort.

gabyx:
I will take a look at this a new PR.
Pushed again, hopefully the tests now work. Forgot to call prepare_target_template_dir always...
Good we have tests :-)

install.sh
gabyx: Ok, but with that in mind, shouldn't the whole code base use  `${...}` in favor of `$...`?  You are right, during regex/replace we ofther forget the `${. *}` case, hm...

rycus86: Yeah, that might be a good change for me, but perhaps it's more personal preference than anything else. Right now I think we have this mixed here and there, so I'm not too bothered.
Maybe if I find some time I'll change all of them and add a test to enforce it. 🤷‍♂️

rycus86:
Looks better, still some core.hooksPath issues it seems?

install.sh
gabyx: Ups that was wrong! Its needed! TEst 067.sh fails otherwise with tests-corehookspath.sh
But of course its needed....

gabyx:
Jeah test-067.sh was the problem. I accidentally removed 
```sh
# Using core.hooksPath implies it applies to all repo's
    if is_single_repo_install && use_core_hookspath; then
        echo "! Cannot use --single and --use-core-hookspath together" >&2
        return 1
    fi
```
which was wrong...

gabyx:
Question: Shouldn't all .sh scripts inside `tests` be executable.
Why is no `.sh` script executable in the repo?

rycus86:
> Question: Shouldn't all .sh scripts inside `tests` be executable.
> Why is no `.sh` script executable in the repo?

I always call them with `sh <something>.sh` so I never bothered to set them I think. The actual steps should also not run by themselves, only through the test runner I guess.

Is it causing any problems for you or just curious?

gabyx:
Nono, I was just thinking, ah ok, tough the reasoning is *all scripts which should not be executable by them self shouldn't be executable*.

Maybe make all `tests-*.sh` scripts executable. though I dont have to write all ways `sh ... ` :)?

rycus86:
Yeah, put up a PR if you have strong feelings about those test runners. :)

rycus86:
Looks like Travis is happy, let me try to review this then I'll merge it. 🎉

gabyx:
Thx!

install.sh
```diff
@@ -177,6 +176,24 @@ legacy_transformations_end() {
...
+############################################################
+# Function to dispatch to all legacy transformations
+#   at the end of a install into a repository
+#
+# Returns:
+#   1 when failed, 0 otherwise
+############################################################
+legacy_transformations_repo_end() {
+
+    git -C "$1" config --local --unset githooks.autoupdate.registered
```
rycus86, line 195: Do we re-register already register repos on the first install/update here?
If not, should this need migration from the old config to the new config?

uninstall.sh
```diff
@@ -186,20 +186,18 @@ uninstall_from_existing_repositories() {
...
-#   `autoupdate.registered` which gets deleted
-#    at the end.
+# Uninstall from all registered repositories
 #
 # Returns: 0
 #####################################################
 uninstall_from_registered_repositories() {
 
-    LIST="$INSTALL_DIR/autoupdate/registered"
+    LIST="$INSTALL_DIR/registered"
```
rycus86, line 195: With this change, we'll leave the previous file alone. Perhaps delete both the old and the new path?

gabyx: Good point

rycus86:
OK, nothing serious, added a few comments for you to look at, please just open another PR if you think there's something to do with them.
Merging this one now.

install.sh
gabyx: We copied the registered list which already contains all reigstered repos so far, and any single install repo will still have old hooks (if it got not hit by the `install into existing`)  meaning it wont get registered... 

Ups we have forgotten a legacy transformation which would ask to search for single repo installs and then making them `non-single`.... Maybe we could add that, or is it not so important?

rycus86: If the update touches them and they are still single then yeah making them non-single sounds sensible to me.

PR #114, opened 2020-06-17
Bugfix: Error if update clone is wrong configured

gabyx:
Also prompt the user what he wants to do, default is no recloning


gabyx:
Small type: Now all tests should work, thanks for merging

rycus86:
> Overall coverage decreased (-1.3%) to 79.54%

We might need to check soon what areas are missing coverage ...

gabyx:
Jeah, probably new introduces stuff, like this PR... sadly...
Thanks for merging, we get some really good progress here :-)

Issue #82, opened 2020-04-16
Global hooks from local directory?


matthijskooijman:
What I'd like to do with githooks, is have a global set of hooks that run for *all* repositories on my system. I'd like to use githooks, rather than just putting the hooks in the global `core.hooksPath` directory, in order to allow *also* running `.githooks` in repositories, and to allow cleanly separating various stuff in multiple files.

It seems there is some support for this through shared hooks, which can be pulled in by a specfic repository using a `.githooks/.shared` file that lists repository urls to pull hooks from. Also, the `githooks.shared` config variable can also add repositories to this list for *all* repositories on my system.

The latter is is *almost* what I want, except that AFAIU this:
 - needs these global hooks to live in a cloneable git repo, while I'd rather just use a global folder (but I can probably work around that with a file:// URL).
 - creates a clone of my global shared repo in the `.githooks` directory of *all* repos on my system, which I think is a bit overkill. Just having a single directory (git repo or just plain directory) that is directly used seems easy to implement and just as powerful?

matthijskooijman:
Hm, it seems I might have misunderstood the documentation somewhat. I thought the shared hooks would be cloned into the repository `.githooks` directory, but the documentation quite clearly says "hese repositories will be checked out into the <install-prefix>/.githooks/shared folder (~/.githooks/shared by default)," (so not sure where my confusion came from).

So, (global) shared hooks are only checked out once, which solves half of my problem. I'd still like to have just a locally managed directory, rather than a git checkout, but that might be easy to add, or workaround :-)

matthijskooijman:
I ended up just doing this:
```
--- a/base-template.sh
+++ b/base-template.sh
@@ -30,6 +30,7 @@ process_git_hook() {
     execute_global_shared_hooks "$@" || return 1
     execute_local_shared_hooks "$@" || return 1
     execute_all_hooks_in "$(pwd)/.githooks" "$@" || return 1
+    execute_all_hooks_in "$HOOK_FOLDER/../global-hooks" "$@" || return 1
 }
 
 #####################################################
```

This just runs all hooks from a global directory of my choice. Here it's relative to the hook folder since I'm running from a git checkout, see #83, but in general it would probably be more useful if the path would be loaded from a git config variable instead.

gabyx:
Maybe we could support such a thing easily:
Lest se what @rycus86 thinks. I think `git config --global githooks.local` would be ideal (analog to `githooks.shared` a separated list of local folders instead of git repos...

[...]

PR #119, opened 2020-09-04
Allow urls and local paths in shared hooks 

gabyx:
- local urls do not get cloned nor updated

See #82. Not testet, first implementation.

Comments highly welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,7 +105,7 @@ For this reason, you can place a `.shared` file inside the `.githooks` repositor
 
 ```shell
 $ git config --global --get githooks.shared
-https://github.com/shared/hooks-python.git,ssh://github.com/shared/hooks-maven.git
+https://github.com/shared/hooks-python.git,git@github.com:shared/repo.git@mybranch
 $ git hooks shared list
 ...
 ```
@@ -114,23 +114,23 @@ The install script offers to set these up for you, but you can do it any time by
 
 Supported entries for shared hooks are:
 
-- **All URLs [Git supports](https://git-scm.com/docs/git-clone#_git_urls)**:
+- **All URLs [Git supports](https://git-scm.com/docs/git-clone#_git_urls)** such as:
 
-  - `ssh://github.com/shared/hooks-maven.git@mybranch`
+  - `ssh://github.com/shared/hooks-maven.git@mybranch` and also the short `scp` form
+     `git@github.com:shared/hooks-maven.git`
   - `git://github.com/shared/hooks-python.git`
   - `file://.*` *see local paths below*
 
   All URLs can include a tag specification syntax at the end like `...@<tag>`, where `<tag>` is a Git tag, branch or commit hash.
-
-- **Local paths** to bare and non-bare repositories:
+- **Local paths** to bare and non-bare repositories such as:
 
   - `/local/path/to/bare-repo.git` (gets used directly)
   - `file:///local/path/to/bare-repo.git@mybranch` (gets cloned internally)
 
   Note that relative paths are relative to the path of the repository executing the hook.
-  These entries are by default forbidden for **local shared hooks** (the `.shared` file) because it makes little sense and is a security risk.
+  These entries are forbidden for **local shared hooks** (`.githooks/.shared` per repository) because it makes little sense and is a security risk.
 
-Shared repositories which are *not local paths or point to a local bare repository* will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. Any other local shared repository will be used directly and not updated nor modified.
+Shared repositories which are *not local paths or point to a local bare repository* will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. Any other local shared repository will be used **directly and will not be updated or modified**.
 
 The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
 
@@ -282,7 +282,7 @@ cd bareRepo
 git hooks install
 # Creates .githooks/trust-all marker for this bare repo
 git hooks trust
-# Automaticall accept changes to all existing and new
+# Automatically accept changes to all existing and new
 # hooks in the current repository.
 git hooks config accept trusted
 # Don't do global automatic updates, since the Githooks update
--- a/README.md
+++ b/README.md
@@ -106,7 +106,7 @@ For this reason, you can place a `.shared` file inside the `.githooks` repositor
 ```shell
 $ git config --global --get githooks.shared
 git@github.com:shared/hooks-python.git,git@github.com:shared/hooks-maven.git
-$ git hooks shared list --with-url
+$ git hooks shared list
 ...
 ```
 
--- a/README.md
+++ b/README.md
@@ -105,12 +105,37 @@ For this reason, you can place a `.shared` file inside the `.githooks` repositor
 
 ```shell
 $ git config --global --get githooks.shared
-git@github.com:shared/hooks-python.git,git@github.com:shared/hooks-maven.git
+https://github.com/shared/hooks-python.git,ssh://github.com/shared/hooks-maven.git
 $ git hooks shared list
 ...
 ```
 
-The install script offers to set these up for you, but you can do it any time by changing the global configuration variable. These repositories will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
+The install script offers to set these up for you, but you can do it any time by changing the global configuration variable.
+
+Supported entries for shared hooks are:
+
+- **All URLs [Git supports](https://git-scm.com/docs/git-clone#_git_urls)**:
+
+  - `ssh://github.com/shared/hooks-maven.git@mybranch`
+  - `git://github.com/shared/hooks-python.git`
+  - `file://.*` *see local paths below*
+
+  All URLs can include a tag specification syntax at the end like `...@<tag>`, where `<tag>` is a Git tag, branch or commit hash.
+
+- **Local paths** to bare and non-bare repositories:
+
+  - `/local/path/to/bare-repo.git` (gets used directly)
+  - `file:///local/path/to/bare-repo.git@mybranch` (gets cloned internally)
+
+  These entries are by default forbidden for **local shared hooks** (the `.shared` file) because it makes little sense.
+  However, you can override this by setting the local/global Git configuration variable `githooks.allowLocalPathsinLocalSharedHooks` to `true`. This is especially useful for setups on a Git server.
+
+  Note that relative paths are relative to the path of the repository executing the hook.
+
+Shared repositories which are *not local paths or point to a local bare repository* will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. Any other local shared repository will be used directly and not updated nor modified.
+
+The layout of these shared repositories is the same as above, with the exception that the hook folders (or files) can be at the project root as well, to avoid the redundant `.githooks` folder.
+
 An additional global configuration parameter `githooks.failOnNonExistingSharedHooks` makes hooks fail with an error if any shared hook configured in `.shared` is missing, meaning `git hooks update` has not yet been called. See `git hooks config [enable|disable] fail-on-non-existing-shared-hooks` in the [command line helper](https://github.com/rycus86/githooks/blob/master/docs/command-line-tool.md) tool documentation for more information.
 Note that shared hooks are automatically updated on clone.
 
--- a/README.md
+++ b/README.md
@@ -127,10 +127,8 @@ Supported entries for shared hooks are:
   - `/local/path/to/bare-repo.git` (gets used directly)
   - `file:///local/path/to/bare-repo.git@mybranch` (gets cloned internally)
 
-  These entries are by default forbidden for **local shared hooks** (the `.shared` file) because it makes little sense.
-  However, you can override this by setting the local/global Git configuration variable `githooks.allowLocalPathsinLocalSharedHooks` to `true`. This is especially useful for setups on a Git server.
-
   Note that relative paths are relative to the path of the repository executing the hook.
+  These entries are by default forbidden for **local shared hooks** (the `.shared` file) because it makes little sense and is a security risk.
 
 Shared repositories which are *not local paths or point to a local bare repository* will be checked out into the `<install-prefix>/.githooks/shared` folder (`~/.githooks/shared` by default), and are updated automatically after a `post-merge` event (typically a `git pull`) on any local repositories. Any other local shared repository will be used directly and not updated nor modified.
 
--- a/base-template.sh
+++ b/base-template.sh
@@ -606,9 +606,9 @@ update_shared_hooks_if_appropriate() {
                 [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
                 echo "! Warning: Local shared hooks contain a local path" >&2
                 echo "  \`$SHARED_REPO\`" >&2
-                echo "  which is forbidden. It will be skipped." >&2
+                echo "  which is forbidden. Update will be skipped." >&2
                 echo ""
-                echo "  You can only have local paths in shared repository defined" >&2
+                echo "  You can only have local paths for shared hooks defined" >&2
                 echo "  in the local or global Git configuration." >&2
                 echo ""
                 echo "  This can be achieved by running" >&2
@@ -687,6 +687,14 @@ execute_shared_hooks() {
             echo "! Local shared hooks contain a local path" >&2
             echo "  \`$SHARED_REPO\`" >&2
             echo "  which is forbidden." >&2
+            echo ""
+            echo "  You can only have local paths in shared hooks defined" >&2
+            echo "  in the local or global Git configuration." >&2
+            echo ""
+            echo "  This can be achieved by running" >&2
+            echo "    \$ git hooks shared add [--local|--global] \"$SHARED_REPO\"" >&2
+            echo "  and deleting it from the \`.shared\` file by" >&2
+            echo "    \$ git hooks shared remove \"$SHARED_REPO\"" >&2
             return 1
 
         fi
--- a/base-template.sh
+++ b/base-template.sh
@@ -109,6 +109,10 @@ set_main_variables() {
     # Global IFS for loops
     IFS_COMMA_NEWLINE=",
 "
+
+    # Fail if the shared root is not available (if enabled)
+    FAIL_ON_NOT_EXISTING_SHARED_HOOK=$(git config --get githooks.failOnNonExistingSharedHooks)
+
     return 0
 }
 
@@ -595,13 +599,10 @@ update_shared_hooks_if_appropriate() {
                 # Non-cloned roots are ignored
                 continue
             elif [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
-                [ "$SHARED_REPO_IS_LOCAL" = "true" ] &&
-                [ "$(git config githooks.allowLocalPathsInLocalSharedHooks 2>/dev/null)" != "true" ]; then
+                [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
                 echo "! Warning: Local shared hooks contain a local path" >&2
                 echo "  \`$SHARED_REPO\`" >&2
-                echo "  which is discouraged. It will be skipped." >&2
-                echo "  To allow this (e.g. on server repositories) you can run:" >&2
-                echo "    \$ git config githooks.allowLocalPathsInLocalSharedHooks \"true\"" >&2
+                echo "  which is forbidden. It will be skipped." >&2
                 continue
             fi
 
@@ -664,23 +665,18 @@ execute_shared_hooks() {
     # split on comma and newline
     IFS="$IFS_COMMA_NEWLINE"
 
-    # Fail if the shared root is not available (if enabled)
-    FAIL_ON_NOT_EXISTING=$(git config --get githooks.failOnNonExistingSharedHooks)
-
     for SHARED_REPO in $SHARED_REPOS_LIST; do
         unset IFS
 
         set_shared_root "$SHARED_REPO"
 
         if [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
-            [ "$SHARED_REPO_IS_LOCAL" = "true" ] &&
-            [ "$(git config githooks.allowLocalPathsInLocalSharedHooks 2>/dev/null)" != "true" ]; then
-            echo "! Warning: Local shared hooks contain a local path" >&2
+            [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+            echo "! Local shared hooks contain a local path" >&2
             echo "  \`$SHARED_REPO\`" >&2
-            echo "  which is discouraged. It will be skipped." >&2
-            echo "  To allow this (e.g. on server repositories) you can run:" >&2
-            echo "    \$ git config githooks.allowLocalPathsInLocalSharedHooks \"true\"" >&2
-            continue
+            echo "  which is forbidden." >&2
+            return 1
+
         fi
 
         if [ ! -d "$SHARED_ROOT" ]; then
@@ -693,7 +689,7 @@ execute_shared_hooks() {
                 echo "    \$ git hooks shared update" >&2
             fi
 
-            if [ "$FAIL_ON_NOT_EXISTING" = "true" ]; then
+            if [ "$FAIL_ON_NOT_EXISTING_SHARED_HOOK" = "true" ]; then
                 return 1
             else
                 echo "  Continuing..." >&2
@@ -712,7 +708,7 @@ execute_shared_hooks() {
                 echo "    \$ git hooks shared purge" >&2
                 echo "    \$ git hooks shared update" >&2
 
-                if [ "$FAIL_ON_NOT_EXISTING" = "true" ]; then
+                if [ "$FAIL_ON_NOT_EXISTING_SHARED_HOOK" = "true" ]; then
                     return 1
                 else
                     echo "  Continuing..." >&2
--- a/base-template.sh
+++ b/base-template.sh
@@ -470,7 +470,6 @@ run_hook_file() {
         # Run as a Shell script
         sh "$HOOK_PATH" "$@"
         return $?
-
     fi
 
     return 0
@@ -493,36 +492,80 @@ process_shared_hooks() {
 }
 
 #####################################################
-# Check is not a supported git clone url and
-#   is treated as local path.
-# Returns:
-#   none
+# Check if `$1` is not a supported git clone url and
+#   is treated as a local path to a repository
+#
+# Returns: 0 if it is a local path, 1 otherwise
 #####################################################
-is_local_url() {
-    if echo "$1" | grep -Eq "[a-zA-Z]+://"; then
+is_local_path() {
+    if echo "$1" | grep -Eq "^[^/]+://"; then
         return 1
     fi
     return 0
 }
 
 #####################################################
-# Sets the SHARED_ROOT and NORMALIZED_NAME
-#   for the shared hook repo url `$1`.
+# Check if url `$1`is a local url, e.g `file://`.
+#
+# Returns: 0 if it is a local url, 1 otherwise
+#####################################################
+is_local_url() {
+    if echo "$1" | grep -Eq "^\s*file://"; then
+        return 0
+    fi
+    return 1
+}
+
+#####################################################
+# Sets the `SHARED_ROOT` for the shared hook repo
+#    url `$1` and sets
+#   `SHARED_REPO_IS_CLONED` to `true` and its
+#   `SHARED_REPO_CLONE_URL` if is needs to get
+#    cloned and `SHARED_REPO_IS_LOCAL` to `true`
+#    if `$1` points to to a local path.
 #
 # Returns:
 #   none
 #####################################################
 set_shared_root() {
-    if ! is_local_url "$1"; then
-        NORMALIZED_NAME=$(echo "$1" |
-            sed -E "s#.*[:/](.+/.+)\\.git#\\1#" |
-            sed -E "s/[^a-zA-Z0-9]/_/g")
-        SHARED_ROOT="$INSTALL_DIR/shared/$NORMALIZED_NAME"
-    else
-        SHARED_ROOT="$1"
-    fi
-}
 
+    SHARED_REPO_CLONE_URL=""
+    SHARED_REPO_CLONE_BRANCH=""
+    SHARED_REPO_IS_LOCAL="false"
+    SHARED_REPO_IS_CLONED="true"
+    DO_SPLIT="true"
+
+    if is_local_path "$1"; then
+        SHARED_REPO_IS_LOCAL="true"
+
+        if is_bare_repo "$1"; then
+            DO_SPLIT="false"
+        else
+            SHARED_REPO_IS_CLONED="false"
+            SHARED_ROOT="$1"
+            return
+        fi
+    elif is_local_url "$1"; then
+        SHARED_REPO_IS_LOCAL="true"
+    fi
+
+    # Here we have now a `<protocol>:/.*(@<tag>)?` or
+    # local bare-repo `<localpath>`
+
+    # Split "...@(.*)"
+    if [ "$DO_SPLIT" = "true" ] && echo "$1" | grep -q "@"; then
+        SHARED_REPO_CLONE_URL="$(echo "$1" | sed -E "s|^(.*)@.*$|\\1|")"
+        SHARED_REPO_CLONE_BRANCH="$(echo "$1" | sed -E "s|^.*@(.*)$|\\1|")"
+    else
+        SHARED_REPO_CLONE_URL="$1"
+        SHARED_REPO_CLONE_BRANCH=""
+    fi
+
+    # Define the shared clone folder
+    CHECKSUM=$(echo "$1" | md5sum | awk "{ print \$1 }")
+    NAME=$(echo "$1" | tail -c 48 | sed -E "s/[^a-zA-Z0-9]/-/g")
+    SHARED_ROOT="$INSTALL_DIR/shared/$CHECKSUM-$NAME"
+}
 #####################################################
 # Update the shared hooks that are on the
 #   ${SHARED_REPOS_LIST} variable.
@@ -546,15 +589,18 @@ update_shared_hooks_if_appropriate() {
         for SHARED_REPO in $SHARED_REPOS_LIST; do
             unset IFS
 
-            if is_local_url "$SHARED_REPO"; then
-                # We only update external urls
-                # which have been cloned
+            set_shared_root "$SHARED_REPO"
+
+            if [ "$SHARED_REPO_IS_CLONED" != "true" ]; then
+                # Non-cloned roots are ignored
+                continue
+            elif [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
+                [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+                echo "! Warning: Local shared hooks contain a local path" >&2
+                echo "  \`$SHARED_REPO\`" >&2
+                echo "  to a local repository which is non-sense." >&2
                 continue
             fi
-
-            mkdir -p "$INSTALL_DIR/shared"
-
-            set_shared_root "$SHARED_REPO"
 
             if [ -d "$SHARED_ROOT/.git" ]; then
                 echo "* Updating shared hooks from: $SHARED_REPO"
@@ -565,9 +611,32 @@ update_shared_hooks_if_appropriate() {
                     echo "$PULL_OUTPUT" >&2
                 fi
             else
-                echo "* Retrieving shared hooks from: $SHARED_REPO"
-                [ -d "$SHARED_ROOT" ] && rm -rf "$SHARED_ROOT"
-                CLONE_OUTPUT=$(git -c core.hooksPath=/dev/null clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
+                echo "* Retrieving shared hooks from: $SHARED_REPO_CLONE_URL"
+
+                [ -d "$SHARED_ROOT" ] &&
+                    rm -rf "$SHARED_ROOT" &&
+                    mkdir -p "$SHARED_ROOT"
+
+                ADD_ARGS=""
+                [ "$SHARED_REPO_IS_LOCAL" != "true" ] && ADD_ARGS="--depth 1"
+
+                if [ -n "$SHARED_REPO_CLONE_BRANCH" ]; then
+                    # shellcheck disable=SC2086
+                    CLONE_OUTPUT=$(git -c core.hooksPath=/dev/null clone \
+                        --single-branch \
+                        --branch "$SHARED_REPO_CLONE_BRANCH" \
+                        $ADD_ARGS \
+                        "$SHARED_REPO_CLONE_URL" \
+                        "$SHARED_ROOT" 2>&1)
+                else
+                    # shellcheck disable=SC2086
+                    CLONE_OUTPUT=$(git -c core.hooksPath=/dev/null clone \
+                        --single-branch \
+                        $ADD_ARGS \
+                        "$SHARED_REPO_CLONE_URL" \
+                        "$SHARED_ROOT" 2>&1)
+                fi
+
                 # shellcheck disable=SC2181
                 if [ $? -ne 0 ]; then
                     echo "! Clone failed, git clone output:" >&2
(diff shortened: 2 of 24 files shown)


gabyx:
We want strict defined behavior, and thats why I go with these rules:

1. all urls in local and global shared hooks which do **not start** with `<protocol>://` are treated as non clonable paths, otherwise they get cloned and the default logic happens. 

For protcols urls `(https?|file|ssh|ftp|git|file)://`
- [ ] a url=`<protocol>://<path>@<tag>`  makes cloning and checking out <tag>... (folder name is the whole url)

For non-clonable paths
- [ ] Relative paths are relative to the **repository**, absolut are **absolut** (TODO)
- [x] we dont clone, but directy use these paths (no changes are made to these path in **any** circumstance, since we are not under control



Windows is not a concern, since everything is wrapped through `bash.exe` anyways and it works :-)!
Paths should be with **forward slashes** (and anything otherwise we define as unsupported, it might work however...), we do it as git does it...

rycus86:
Is this code shared with the in-repo shared hooks file handling? I'd like to make sure that I can't commit a shared hook config into a repo we both use that runs something malicious from your machine's local file system. Not sure yet how I would exploit it, but let's make sure we can't. Thanks!

base-template.sh
```diff
@@ -491,6 +491,19 @@ process_shared_hooks() {
...
 }
 
+#####################################################
+# Check is not a supported git clone url and
+#   is treated as local path.
+# Returns:
+#   none
+#####################################################
+is_local_url() {
+    if echo "$1" | grep -Eq "(https?|file|ssh|ftp|git|file)://"; then
```
rycus86, line 501: Maybe simply check the protocol separator
```suggestion
    if echo "$1" | grep -Eq "[a-zA-Z]+://"; then
```

rycus86: Btw, this might be fun to test with our existing tests... Though I guess we can just switch them over to `file://`

gabyx:
Ah jeah, your are right :-) may be we should only consider `global` shared hooks. I did not check that yet :-) Thx!

gabyx:
> Is this code shared with the in-repo shared hooks file handling? I'd like to make sure that I can't commit a shared hook config into a repo we both use that runs something malicious from your machine's local file system. Not sure yet how I would exploit it, but let's make sure we can't. Thanks!

May be you can exploit it by simply putting `.` as a url.
Meaning we execute the hooks in this repo again -> endless recursion lool, dont know...

rycus86:
New logic seems OK to me. 👍
Can you add some tests please?

gabyx:
Jeah, will do that later!

base-template.sh
```diff
@@ -484,13 +484,27 @@ run_hook_file() {
...
 }
 
+#####################################################
+# Check is not a supported git clone url and
+#   is treated as local path.
+# Returns:
+#   none
+#####################################################
+is_local_url() {
+    if echo "$1" | grep -Eq "[a-zA-Z]+://"; then
```
rycus86, line 502: Now that I look at this again, I suppose it's still too strict. Perhaps something like this?
```suggestion
    if echo "$1" | grep -Eq "[^/]+://"; then
```

gabyx:
Added tests and finished for review. Seems to work fine.
- Warning for execution
- Warning for adding
for adding local repo url to  **local** list. (Global ist allowed)

base-template.sh
gabyx: Why you mean? :-) Is there such a obscure protocol string?

gabyx:
**Note**: I changed the name how clones shared repos are stored in `.githooks/shared/xxx`
I needed this, since I think its good to have the full URL there... full distinguising between

- https://bla.git@master2
- https://bla.git@master
- http://bla.git@master2

base-template.sh
rycus86: `socks5://` for example, not sure it's something you can actually use with Git, but I have no idea what are all the things you can give to it. 🙂

base-template.sh
```diff
@@ -484,27 +484,76 @@ run_hook_file() {
...
+    SHARED_REPO_IS_LOCAL="false"
+    SHARED_REPO_IS_CLONED="true"
+
+    # Split "...@(.*)"
+    if echo "$1" | grep -q "@"; then
+        URL="$(echo "$1" | sed -E "s|^(.*)@.*$|\\1|")"
+        BRANCH="$(echo "$1" | sed -E "s|^.*@(.*)$|\\1|")"
+    else
+        URL="$1"
+        BRANCH="master"
```
rycus86, line 532: I think this would be whatever you set as default branch on the server, right?

tests/step-005.sh
```diff
@@ -15,14 +15,14 @@ cd /tmp/shared/hooks-005.git &&
 if echo "$EXTRA_INSTALL_ARGS" | grep -q "use-core-hookspath"; then
     echo 'n
 y
-/tmp/shared/hooks-005.git
+file:///tmp/shared/hooks-005.git
```
rycus86, line 18: Just checking that these should still work without changes, right?
E.g. we don't need users to set up existing shared hooks again?

rycus86: Or this is the new check where we don't allow local paths as shared repos unless  you make it local due to security concerns?
If that's the case, then it's fine.

rycus86:
> **Note**: I changed the name how clones shared repos are stored in `.githooks/shared/xxx`
> I needed this, since I think its good to have the full URL there... full distinguising between
> 
> * https://bla.git@master2
> * https://bla.git@master
> * http://bla.git@master2

Can you please put it here an example of absolute paths these would become in the checked out shared repo on the filesystem? Without normalization can't they contain characters that are invalid on some OSes, like Windows and `:` for example?

rycus86:
FYI there seem to be test failures but the Travis integration keeps breaking and doesn't show up here anymore. 😔
https://travis-ci.org/github/rycus86/githooks

base-template.sh
gabyx: https://git-scm.com/docs/git-clone#_git_urls

base-template.sh
gabyx: ah, ok, so we set it to empty and then dont hand that parameter to the clone function

tests/step-005.sh
gabyx: Correction: Unless you make it global, local is prohibited.
Since I changed the folder nameing, **we need to tell the user to setup shared hooks again**: Do we need post install logic for this update? Or just a warning if the user updates from a earlier commit then this date we show it...

gabyx:
> Can you please put it here an example of absolute paths these would become in the checked out shared repo on the filesystem? Without normalization can't they contain characters that are invalid on some OSes, like Windows and `:` for example?

* https://bla.git@master2 -> "https-bla.git@master2"
* https://bla.git@master -> "https-bla.git@master"
* http://bla.git@master2 -> "http-bla.git@master"

Correct, we replace all special characters as you did before!

gabyx:
I will check the tests later.

base-template.sh
rycus86: On that page it says:
> When Git doesn’t know how to handle a certain transport protocol, it attempts to use the remote-<transport> remote helper, if one exists.

My reading of this is that you can have `myF4ncyTr4sp0rt://` URLs and Git will use a helper to handle that.

gabyx: Ah good, dint read it carefully :-). Ok lets do this :-)

tests/step-005.sh
rycus86: So if the user added local shared repos that are in the config (but not checked in through the .shared file) those would be nice to either keep working or auto-upgrade or something. If there are absolute paths without a protocol in the .shared file in a repo, I'm OK to have those broken then they'll see the warning you added.

rycus86:
> 
> > Can you please put it here an example of absolute paths these would become in the checked out shared repo on the filesystem? Without normalization can't they contain characters that are invalid on some OSes, like Windows and `:` for example?
> 
> * https://bla.git@master2 -> "https-bla.git@master2"
> * https://bla.git@master -> "https-bla.git@master"
> * http://bla.git@master2 -> "http-bla.git@master"
> 
> Correct, we replace all special characters as you did before!

Excellent, that should work then hopefully. Could you perhaps add a test to showcase this please? (if not too difficult)

gabyx:
@rycus86 : We have another covarage kcov to-old-git issue -> good news: i think I finally fixed kcov v36 :-)
See next commit:
- it needs a user in the docker container
- I (for now, other solution?) added the kvov run directly into the image building
- copy the output from the image

rycus86:
> @rycus86 : We have another covarage kcov to-old-git issue -> good news: i think I finally fixed kcov v36 :-)
> See next commit:
> - it needs a user in the docker container
> - I (for now, other solution?) added the kvov run directly into the image building
> - copy the output from the image

Ooh, exciting! I'll have a look later today. Thanks!

gabyx:
> With that fixes in coverage.sh it runs through and I report towards "~/.githooks/release".
Not sure, but in any case if we use `var/lib/githooks` all release clone calls in `~/.githooks/release` are anyway not counted, aren't they? 
Not sure but I dont think kcov has support for merging   same file names? `--merge`

rycus86:
> > With that fixes in coverage.sh it runs through and I report towards "~/.githooks/release".
> 
> Not sure, but in any case if we use `var/lib/githooks` all release clone calls in `~/.githooks/release` are anyway not counted, aren't they?
> Not sure but I dont think kcov has support for merging same file names? `--merge` ? Hm...
> So one should maybe rewrite the tests by having the direct wrapper calls (not install if you really dont want it) but just place them in the same directory :-) -> `~/.githooks/release/...` at the beginngin of the test.
> Or hardlinking towards `/var/lib/githooks/....` might also be a nice option...

https://coveralls.io/builds/33289402 seems to be fine for me, is there anything missing from it? Looks like `cli.sh` is not there, not sure if that's related to this change? Otherwise I wouldn't be worried where we get the data from as long as it's there.

base-template.sh
```diff
@@ -484,27 +483,73 @@ run_hook_file() {
 #   1 in case a hook fails, 0 otherwise
 #####################################################
 process_shared_hooks() {
-    SHARED_REPOS_LIST="$1"
-    shift
+    SHARED_HOOKS_TYPE="$1"
+    SHARED_REPOS_LIST="$2"
+    shift 2
```
rycus86, line 488: Is this POSIX?
If not (or unsure like I am) perhaps do
```
SHARED_HOOKS_TYPE="$1"
shift
SHARED_REPOS_LIST="$1"
shift
```

base-template.sh
```diff
@@ -527,10 +572,23 @@ update_shared_hooks_if_appropriate() {
...
-            mkdir -p "$INSTALL_DIR/shared"
 
             set_shared_root "$SHARED_REPO"
 
+            if [ "$SHARED_REPO_IS_CLONED" != "true" ]; then
+                # Non-cloned roots are ignored
+                continue
+            elif [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
+                [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+                echo "! Warning: Local shared hooks contain an url" >&2
```
rycus86, line 583: Please use `URL` everywhere instead of `url`.
Also I believe it's `a URL` rather than `an URL`, please. :bow:

base-template.sh
```diff
@@ -540,9 +598,24 @@ update_shared_hooks_if_appropriate() {
...
-                CLONE_OUTPUT=$(git -c core.hooksPath=/dev/null clone "$SHARED_REPO" "$SHARED_ROOT" 2>&1)
+
+                if [ -n "$SHARED_REPO_CLONE_BRANCH" ]; then
+                    CLONE_OUTPUT=$(git -c core.hooksPath=/dev/null clone \
+                        --single-branch \
+                        --branch "$SHARED_REPO_CLONE_BRANCH" \
+                        "$SHARED_REPO_CLONE_URL" \
+                        "$SHARED_ROOT" 2>&1)
+                else
+                    # shellcheck disable=SC2090,SC2086
```
rycus86, line 612: Why do we need to disable these? (if we don't have to above)

base-template.sh
```diff
@@ -575,6 +648,14 @@ execute_shared_hooks() {
 
         set_shared_root "$SHARED_REPO"
 
+        if [ "$SHARED_HOOKS_TYPE" = "--local" ] && [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+            echo "! Warning: Local shared hooks contain an url" >&2
```
rycus86, line 652:
```suggestion
            echo "! Warning: Local shared hooks contain a URL" >&2
```

base-template.sh
```diff
@@ -588,22 +669,23 @@ execute_shared_hooks() {
...
-            else
-                echo "  Continuing..." >&2
-                continue
+        if [ "$SHARED_REPO_IS_CLONED" = "true" ]; then
+            # Note: GIT_DIR might be set (?bug?) (actually the case for post-checkout hook)
+            # which means we really need a `-f` to sepcify the actual config!
+            REMOTE_URL=$(git -C "$SHARED_ROOT" config -f "$SHARED_ROOT/.git/config" --get remote.origin.url)
+            if ! echo "$SHARED_REPOS_LIST" | grep -q "$REMOTE_URL"; then
+                echo "! Failed to execute shared hooks in $SHARED_REPO" >&2
+                echo "  The remote url \`$REMOTE_URL\` is different." >&2
```
rycus86, line 678:
```suggestion
                echo "  The remote URL \`$REMOTE_URL\` is different." >&2
```

cli.sh
```diff
@@ -950,7 +949,16 @@ add_shared_hook_repo() {
 
     else
         if ! is_running_in_git_repo_root; then
-            echo "The current directory ($(pwd)) does not seem to be the root of a Git repository!"
+            echo "! The current directory ($(pwd)) does not
+                  seem to be the root of a Git repository!" >&2
```
rycus86, line 953: Should this be another `echo` here?

cli.sh
```diff
@@ -950,7 +949,16 @@ add_shared_hook_repo() {
...
+            exit 1
+        fi
+
+        if is_local_url "$SHARED_REPO_URL"; then
+            echo "! Adding an local url:" >&2
```
rycus86, line 958:
```suggestion
            echo "! Adding a local URL:" >&2
```

rycus86: What is a local URL btw? Is that an absolute path, or in other words NOT a URL?

cli.sh
```diff
@@ -1163,22 +1167,26 @@ list_shared_hook_repos() {
...
+                    if [ -d "$SHARED_ROOT/.git" ]; then
                         LIST_ITEM_STATE="active"
                     else
                         LIST_ITEM_STATE="invalid"
                     fi
-                else
-                    LIST_ITEM_STATE="pending"
                 fi
 
-                if [ -n "$LIST_WITH_URL" ]; then
```
rycus86, line 1176: Why did you remove this here?
Since we're normalizing the name, I think it might be good if we could continue to print out the actual source of where these came from.
What do you think?

tests/coverage.sh
```diff
@@ -10,11 +10,11 @@ fi
 
 # Build a Docker image on top of kcov with our scripts
 cat <<EOF | docker build --force-rm -t githooks:coverage -f - .
-FROM kcov/kcov:v33
+FROM kcov/kcov:v36
 
 RUN echo 'deb http://deb.debian.org/debian stretch main' >> /etc/apt/sources.list \
     && (apt-get update || true) \
-    && apt-get install -y git
+    && apt-get install -y git python
```
rycus86, line 17: What needs Python here?

tests/step-005.sh
```diff
@@ -15,14 +14,14 @@ cd /tmp/shared/hooks-005.git &&
...
 ' | sh /var/lib/githooks/install.sh || exit 1
 
 else
     echo 'n
 n
 y
-/tmp/shared/hooks-005.git
+file:///tmp/shared/hooks-005.git
```
rycus86, line 24: What's the default shared hook type? Is it local?
If that's the case, some of these don't have to change, right?
It would be good to have a good mix of both local/global URL/non-URL for coverage.

tests/test-opensuse.sh
```diff
@@ -3,8 +3,8 @@
 TEST_DIR=$(dirname "$0")
 
 cat <<EOF | docker build --force-rm -t githooks:opensuse-base -
-FROM opensuse/amd64
-RUN zypper install -y git
+FROM opensuse/tumbleweed
```
rycus86, line 6: Why did this need to change?

base-template.sh
gabyx: Its Posix yes https://pubs.opengroup.org/onlinepubs/009695399/utilities/shift.html

base-template.sh
gabyx: Ups left over

cli.sh
gabyx: Autsch.... jep

cli.sh
gabyx: should be `Adding a local path` yes

cli.sh
gabyx: I do that be default now. Before we showed the weird mangled name (NORMALIZED_NAME, removed this) , which is any way a bit strange.
So because the user does not need to know this, I print the URL/path he entered in the list, which is also directly what we used to clone, that all these repos are in `.githooks/shared/<checksum>-40charsAbreviation` is internal stuf...

gabyx (commented):
Way too old git, with bugs...

tests/step-005.sh
gabyx: if no protocol  its treated as local.
I though about, only tests `non-protocol` stuff (local paths)  in step-117, to better distinguish this feature from the current implementation. I would leave the original tests with `file://` thats also the most common usecase -> with protocol

tests/coverage.sh
gabyx: Not sure, didnt check, the container did not build without...

tests/step-005.sh
gabyx: > https://coveralls.io/builds/33289402 seems to be fine for me, is there anything missing from it? Looks like `cli.sh` is not there, not sure if that's related to this change? Otherwise I wouldn't be worried where we get the data from as long as it's there.

Hm... thats utter strange, why is cli.sh not there? I though we dispatch to the release folder for the cli as well ?
Hmm have to check that

gabyx:
> https://coveralls.io/builds/33289402 seems to be fine for me, is there anything missing from it? Looks like `cli.sh` is not there, not sure if that's related to this change? Otherwise I wouldn't be worried where we get the data from as long as it's there.

Hm... thats utter strange, why is cli.sh not there? I though we dispatch to the release folder for the cli as well ?
Hmm have to check that

gabyx:
Do you know how kcov can track invocations through `git commit` which triggers base-template.sh.
Isn't that whats happening now? This is magic?

gabyx:
Its really sad:

- `sh "$HOME/.githooks/release/cli.sh" version` **doesnt** trigger the coverage
- `sh "/home/coverage/.githooks/release/cli.sh" version` does trigger the coverage

So in the last commit I refactored the tests a bit.

All tests with direct template execution:
- first copy the scripts to `~/.githooks/release/`  to use them from there
All tests which use an install:
- all `/var/lib/githooks/cli.sh` invocation changed to `git hooks` invocation which get replaced in `coverage.sh` to `~/.githooks/release/cli.sh`
- all `/var/lib/githooks/base-template-wrapper.sh` to `~/.githooks/release/base-template-wrapper.sh`

image

gabyx:
On the way of improving the coverage a bit more: Need some more replacements :-)

gabyx:
FYI: the coverage dropped, because there is something wrong, it does nto include for example

```shell
echo "
asdasd
asd
asd
"
```
Thats why the rating dropped. May be we can reformat (sed) these lines into separate echo just for coverage...
Kcov and Bash sucks, Bash has this PS4 security fixes and probably other issues which kcov is not aware. Maybe we need to build the master brach and see what we get. I am guessing if there is a better coverage tool hm...

rycus86:
If you find anything else that can do Shell coverage, I'm happy to give it a try. When I set up the project, this was the one I found working.

It might be good if you could put up a separate PR just for the kcov upgrade but no other functionality change, then we'd get that into master and we'd get a more relevant picture of coverage changes in this (and future) PRs.

Would that be too much work? Tell me if it is, but hoping now that you know how to get it working, it's easy to redo on top of master.

gabyx:
I can do that, its a bit work, but may be yes, better.

gabyx:
By the way: 
echo replacements for kcov

```shell
echo "<foo>
<bar>
echo \"
line 1
line 2
\"" | sed -E '/echo "$/,/^"/{  s/echo "/echo ""/ ; s/^"$/echo ""/  ;  /"/! { s/(.*)/echo "\1"/ } }'
```

rycus86:
> I can do that, its a bit work, but may be yes, better.

Let me know if it too much, then I can try to extract it from this PR if I ever find some time. 🙂

rycus86:
> By the way:
> echo replacements for kcov
> 
> ```shell
> echo "asdasd
> asdasd
> echo \"
> line 1
> line 2
> \"" | sed -E '/echo "$/,/^"/{  s/echo "/echo ""/ ; s/^"$/echo ""/  ;  /"/! { s/(.*)/echo "\1"/ } }'
> ```

That messes with the line numbers though, right?
Maybe not a huge issue, but also lower coverage is fine if we know it's just due to echos and not because of genuinely uncovered lines.

gabyx:
No I think, it want:
```shell
asdasd
asdasd
echo ""
echo "line 1"
echo "line 2"
echo ""
```

gabyx:
No coverage stuff inside, this PR.

tests/step-005.sh
gabyx: Ok, I ll see if we can setup a legacy transform in the install.sh

gabyx:
Here, sorry for the long mess about the Coverage above, again the PR with squased commits.

Todos: From the discussion:
- [ ] if somebody has local paths in `--global githooks.shared` (without a protocol) we should insert during install `file://` because thats what the previous behavior was, but only when upgrading from a commit before a commit in this PR.
- [ ] Maybe also deprecate `githooks.cloneBranch` and incorporate it into `githooks.cloneUrl`. (Some url handling as shared urls)

base-template.sh
```diff
@@ -484,27 +483,73 @@ run_hook_file() {
...
-    SHARED_ROOT="$INSTALL_DIR/shared/$NORMALIZED_NAME"
-}
 
+    SHARED_REPO_CLONE_URL=""
+    SHARED_REPO_CLONE_BRANCH=""
+    SHARED_REPO_IS_LOCAL="false"
+    SHARED_REPO_IS_CLONED="true"
+
+    # Split "...@(.*)"
+    if echo "$1" | grep -q "@"; then
```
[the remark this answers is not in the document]
matthijskooijman: Hm, giving this some more thought, *if* the part before the @ is a path to a bare git repo, the part after the @ could be interpreted as a branch, but if not, the entire thing should be interpreted as a path (include the @ and what's after it). This is not entirely trivial to check, so I wonder if maybe the `@` syntax should just be limited to just `foo://` style urls and forbidden on local paths (whether they point to a bare git repo or not)? If you need a local repo with a custom branch, one can always just use `file://`?

matthijskooijman (commented):
I left some comments inline.

One overall question I have, is what things are supported exactly? I can see six things:
 1. A path to a local bare git repo
 2. A git:// or other url to a remote git repo
 3. A file:// url to a local bare git repo
 4. A local path to a bare git repo
 5. A local path to a git checkout
 6. A local path to an arbitrary directory with hooks

Before this PR, I think 1-4 were supported. All of these result in a git clone to be made into the `shared` directory and hooks to be ran from there. I think this PR adds support for 5. I would like to also see support for 6., as that's my original usecase (I think this is a matter of dropping a check, as I commented inline, which would merge 5. into 6.).

Note that case 3. (file:// urls) is not currently specially handled, but AFAICS the security implications are identical to case 4., which is forbidden for local config? So maybe this should be specially checked? Though it is always hard to be complete, who knows what extra git protocols will appear later that are effectively also local? Also, I wonder, is there really any special significance to referring to a local directory or bare git repo? In the end, the security problem does not occur when cloning, it occurs when running a hook from an arbitrary local directory. And that is already interactively verified based on checksums, so I wonder whether this security check needs to exist at all? If we're worried about cloning from local paths as well as executing hooks, I wonder if we should be worried about cloning from remote urls and maybe ask confirmation before any clone?

I was going to suggest to maybe drop support for 4., hoping that would simplify code by making the `SHARED_REPO_IS_LOCAL` (e.g. is local path, so 3-6)  and `SHARED_REPO_IS_CLONED` (e.g. needs to be cloned, so 1-4) mutually exclusive, but the previous point would make this point invalid and possibly moot.

base-template.sh
```diff
@@ -484,27 +483,73 @@ run_hook_file() {
...
 
 #####################################################
-# Sets the SHARED_ROOT and NORMALIZED_NAME
-#   for the shared hook repo url `$1`.
+# Check if `$1` is not a supported git clone url and
+#   is treated as a local path to a repository
+# Returns:
+#   none
+#####################################################
+is_local_url() {
```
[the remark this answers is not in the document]
gabyx: is_local_path, is better!

base-template.sh
```diff
@@ -484,27 +483,73 @@ run_hook_file() {
...
+    if echo "$1" | grep -Eqv "^[^/]+://"; then
+        return 0
```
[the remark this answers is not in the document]
gabyx: Better an if and a explizit mapping to 0 and 1

base-template.sh
```diff
@@ -575,6 +654,14 @@ execute_shared_hooks() {
...
 
+        if [ "$SHARED_HOOKS_TYPE" = "--local" ] && [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+            echo "! Warning: Local shared hooks contain a local path" >&2
+            echo "  \`$SHARED_REPO\`" >&2
+            echo "  to a local repository which is ignored" >&2
+            echo "  due to security risks!" >&2
+            continue
+        fi
+
         if [ ! -f "$SHARED_ROOT/.git/config" ]; then
```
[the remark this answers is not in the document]
gabyx: Ahh, nice, it might be a non .git directory, ahh, jeah. we can support that as well I think.

gabyx:
Supported is the following: See https://github.com/rycus86/githooks/pull/119#issuecomment-687400696

gabyx:
[x] A path to a local bare git repo (**only global shared hooks**)
[x] A git:// or other url to a remote git repo
[x] A file:// url to a local bare git repo (ups I missed that one!, this shouldn't be allowed for local shared hooks, which now it is)
[x] A local path to a bare git repo (**only global shared hooks**)
[x] A local path to a git checkout (**only global shared hooks**)
[x] A local path to an arbitrary directory with hooks (**only global shared hooks**)

gabyx:
Can we discuss the missing point 
- [x] A file:// url to a local bare git repo (ups I missed that one!, this shouldn't be allowed for local shared hooks, which now it is)

That should classify as local path as well I think... In that way all tests will probably now fail. Because we can't add `file:///tmp/shared` to the local shared hooks

base-template.sh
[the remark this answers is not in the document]
gabyx: The @ style syntax as in the last commit seen (hopefully) is limited to
- procotol urls
- local urls which are bare git repos

base-template.sh
```diff
@@ -527,10 +572,23 @@ update_shared_hooks_if_appropriate() {
...
+            elif [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
+                [ "$SHARED_REPO_IS_LOCAL" = "true" ]; then
+                echo "! Warning: Local shared hooks contain a local path" >&2
+                echo "  \`$SHARED_REPO\`" >&2
+                echo "  to a local repository which is ignored" >&2
+                echo "  due to security risks!" >&2
+                continue
+            fi
+
+            mkdir -p "$INSTALL_DIR/shared"
```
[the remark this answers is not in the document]
gabyx: Good point

base-template.sh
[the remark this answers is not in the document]
gabyx: Your are right. I change it to protocol only.

gabyx:
>  And that is already interactively verified based on checksums, so I wonder whether this security check needs to exist at all? If we're worried about cloning from local paths as well as executing hooks, I wonder if we should be worried about cloning from remote urls and maybe ask confirmation before any clone?

I totally agree, I was thinking the same when implementing this: Maybe we should only do this:
- local paths dont make sense in local hooks (file:// and local paths)  `.githooks/.shared` -> therefore disallow (for the tests, we turn that off where approppriate)

gabyx:
The new changes include all of the following: `test-117.sh` is especially for local stuff...

- [x] A path to a local bare git repo (**only global shared hooks**)
- [x] A git:// or other url to a remote git repo
- [x] A file:// url to a local bare git repo (**only global shared hooks**)
- [x] A local path to a bare git repo (**only global shared hooks**)
- [x] A local path to a git checkout (**only global shared hooks**)
- [x] A local path to an arbitrary directory with hooks (**only global shared hooks**)

We mock certain tests with `git config --global githooks.testingTreatFileProtocolAsRemote "true"`
to pass gracefully with `file://` simulating a remote url. Test 117 doesn't have this and checks if rejection properly works,

gabyx:
I think we need an option to allow for all type of urls/paths in **local shared hooks**. Scenario for hooks in repos on a server (gitea for example) is that you probably really want to setup local urls/paths in

- server-repo/.githooks/.shared 

So I suggest an option to allow this as opt-in:

```shell
git config githooks.allowLocalPathsInLocalSharedHooks
```
We can set that `--global` or per repo wise, passing through the rejection.

matthijskooijman:
IIUC, you're saying that you might want to do this in the `.githooks` directory of a *bare* repository, right?

If so, then maybe the "global/local" distinction is not accurate? From a security standpoint we feel (I'm not 100% convinced, but I'll roll with this assumption for now) that you should not be trusting contents that is stored *in* the repository itself. Iow, anything that can be changed by a push to the repo, should be potentially untrusted and thus not be allowed to refer to a local path. Aside from trust issues, it would not actually make sense to let the repository contents refer to a local path, since there isn't really any way to guarantee such a local path would be available on every machine that potentially has a clone of this repo.

However, from this perspective, a `.githooks` directory in a bare repository, which AFAICS would be manually created and managed by the server administrator and cannot be accessed through git itself, should be in the same "trusted" category as global configuration (and in a different category than `.githooks` in a non-bare repository, which *can* be accessed through git itself).

gabyx:
@matthijskooijman: totally correct. Because this `.githooks` folder is manually managed on the server and cannot be changed thats why I simply introduced a opt in `githooks.allowLocalPathsInLocalSharedHooks`. 

p.s: I think we cannot distinguish really between a server setup or not. Meaning we are dealing with a manually manager bare repo on the server.....

matthijskooijman:
> p.s: I think we cannot distinguish really between a server setup or not. Meaning we are dealing with a manually managed bare repo on the server.....

Why not? Isn't the characteristic to check "Git repo is bare or not"? If it is bare, the contents of the directory is not managed by git, if it is non-bare, the contents *is* managed by (and controllable through) git?

gabyx:
Hm.. That migth be it. I was not sure, about. Whats about bare repos on your machine... Hm doesn't matter. so I think we can maybe get away with the new config and just pass the rejection when its a bare repo...

rycus86:
> I think we need an option to allow for all type of urls/paths in **local shared hooks**. Scenario for hooks in repos on a server (gitea for example) is that you probably really want to setup local urls/paths in
> 
> * server-repo/.githooks/.shared
> 
> So I suggest an option to allow this as opt-in:
> 
> ```shell
> git config githooks.allowLocalPathsInLocalSharedHooks
> ```
> 
> We can set that `--global` or per repo wise, passing through the rejection.

If this is an admin managed server, can't you just set up global hooks with the cli pointing to those shared hooks, rather than using a .shared file that is checked into the repo? I'm trying to understand why do we need this workaround, and I feel like I'm missing something.

gabyx:
That is possible for sure, but it would be good to have control over which repos habe which shared hooks. So local shared hooks are the way to go, which work already in bare repos.

Von meinem iPhone gesendet

rycus86:
Ah right, I thought we have a "set shared repos for this repository only" command, but I see now that only exists for the global setting. Would it work if we could set that config per repo too (not only globally)? I'd prefer that over checking in a .shared file pointing to local paths that really only make sense on the server in your use case (if I understand it correctly) and that way we could avoid this new feature toggle.

What do you think?

gabyx:
hi rycus,
Ah this exist? :) I never seen this: Can you point me at a line :). You can set global shared hooks for certain repos only? I thought we check always —global on githooks.shared?

Von meinem iPhone gesendet

rycus86:
Yeah, it might be global only right now, I'm wondering if we can also make it local for things like what you want it for.

gabyx:
I think instead of checking —global we can easily just leave it out and support that feature I want, that way. Thx @rycus. That option is exactly what I want. So we can forget to expose the allowLocalPathsFor... option, and I need to introduce the sed replacement again to pass file:// in certain tests. So we can do that in seperate PR :)!


Von meinem iPhone gesendet

gabyx:
I reverted the option change:
- disallow local paths in .shared file.
- no changes related to adding --local/--global git config for githooks.shared

If tests pases, this shoudld be ready to final review and merge.

gabyx:
@rycus86 : 

- I added an `legacy_transform` in install which changes all local paths to `file://` etc
image
Tested it: should work :crossed_fingers: 

- I also added `--internal-updated-from $GITHOOKS_CLONE_UPDATED_FROM_COMMIT` to the `install.sh` which was forgotten in a earlier PR, makeing it unable to react on changes in the future. Now this should be fixed.

Can we sooner or later merge this, would be great.

README.md
```diff
@@ -105,12 +105,35 @@ For this reason, you can place a `.shared` file inside the `.githooks` repositor
 
 ```shell
 $ git config --global --get githooks.shared
-git@github.com:shared/hooks-python.git,git@github.com:shared/hooks-maven.git
```
rycus86, line 108: Are these actually unsupported now?

base-template.sh
```diff
@@ -567,46 +665,58 @@ execute_shared_hooks() {
...
     for SHARED_REPO in $SHARED_REPOS_LIST; do
         unset IFS
 
         set_shared_root "$SHARED_REPO"
 
-        if [ ! -f "$SHARED_ROOT/.git/config" ]; then
-            echo "! Failed to execute shared hooks in $SHARED_REPO" >&2
-            echo "  It is not available. To fix, run:" >&2
-            echo "    \$ git hooks shared update" >&2
+        if [ "$SHARED_HOOKS_TYPE" = "--local" ] &&
```
rycus86, line 673: It might be good to suggest how to fix the issue on each of these failure return branches.
Something like, if it's forbidden remove it from the .shared file, if it doesn't exist, remove it from this config, etc.
It would be good to let people self-manage the fix for these.

install.sh
```diff
@@ -192,11 +192,61 @@ legacy_transformations_end() {
...
+
+            git -C "$1" config githooks.failOnNonExistingSharedHooks true >/dev/null 2>&1
+
+            # Add to all non-URLS a `file://` (https://www.gnu.org/software/sed/manual/sed.html#Branching-and-flow-control)
+            # Explanation:
+            # - select first all non-comment lines
+            # - then select the ones which are not starting with <protocol>://
+            # - then adding file:// to them...
+            TEMP=$(mktemp)
+            sed -E '/^[^#].+$/ { /[^\/]+:\/\//! { s/^/file:\/\//g } }' <"$SHARED_FILE" >"$TEMP" &&
```
rycus86, line 242: :see_no_evil: 
This code is a bit complicated, and as a user I can't say I'd filly understand what's happening and what I need to do.
Can we not just put a `#` at the start of those paths and tell the user we disabled the unsupported shared repos (perhaps after a prompt) ?

install.sh
```diff
@@ -267,6 +317,11 @@ parse_command_line_arguments() {
             INTERNAL_INSTALL="true"
         elif [ "$p" = "--internal-postupdate" ]; then
             INTERNAL_POSTUPDATE="true"
+        elif [ "$p" = "--internal-updated-from" ]; then
+            : # nothing to do here
+        elif [ "$prev_p" = "--internal-updated-from" ] && (echo "$p" | grep -qvE '^\-\-.*'); then
+            # INTERNAL_UPDATED_FROM_COMMIT="$p" # not yet used !
```
rycus86, line 323: Why do we need this here if it's unused?

tests/coverage.sh
```diff
@@ -59,7 +59,9 @@ RUN \\
     sed -i -E 's|HOOK_NAME=(.*)|HOOK_NAME="\${HOOK_NAME:-\1}"|' /var/lib/githooks/base-template.sh && \\
     sed -i 's|ACCEPT_CHANGES=|ACCEPT_CHANGES=\${ACCEPT_CHANGES}|' /var/lib/githooks/base-template.sh && \\
     sed -i 's%read -r "\$VARIABLE"%eval "\$VARIABLE=\\\\\$\$(eval echo "\\\\\$VARIABLE")" # disabled for tests: read -r "\$VARIABLE"%' /var/lib/githooks/base-template.sh && \\
-    sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh
+    sed -i -E 's|GITHOOKS_CLONE_URL="http.*"|GITHOOKS_CLONE_URL="/var/lib/githooks"|' /var/lib/githooks/cli.sh /var/lib/githooks/base-template.sh /var/lib/githooks/install.sh && \\
+# Conditionally allow file:// for local shared hooks simulating http:// protocol
```
rycus86, line 63: This flag is to allow `file://` references in the .shared file? (but not local paths?)

README.md
gabyx: Yes, I dont understand these urls? When I check with Git Url Doc I cannot find an example?

tests/step-016.sh
```diff
@@ -25,8 +27,8 @@ mkdir -p /tmp/test16 && cd /tmp/test16 || exit 1
 git init || exit 1
 
 mkdir -p .githooks &&
-    git config --global githooks.shared '/tmp/shared/hooks-016-a.git' &&
-    echo '/tmp/shared/hooks-016-b.git' >.githooks/.shared &&
+    git config --global githooks.shared 'file:///tmp/shared/hooks-016-a.git' &&
```
rycus86, line 30: The one on the git config should still work with the local path, right?
Can you please change those back so we get coverage on that code path as well?

install.sh
gabyx: Fair enough, but the question still remains, why are these now unsupported... Any suggestion on the warning message?
This fix, turns them into `file://` which is reproduces **exactly** the same behavior for users after updating. Such that their share repos continue to execute.

rycus86: Hm, but how of we don't support local paths there anymore (ecen prefixed with file://), is that not the case?

install.sh
gabyx: Because: We have no chance of getting the commit SHA from the update when a user updated from an old version which doesnt hand over the SHA to the **updated** install.sh, and such we also dont have it when we introduce a change. So from now on we have this SHA hash, also we dont use it yet. Hopefully we never need to really use it =)

README.md
rycus86: I think this is the ssh clone URL you get from GitHub, is it not?

tests/coverage.sh
gabyx: Correct: This is a unit test flag.

install.sh
rycus86: Fair enough. I don't think it hurts to have it, though also unsure how we could use it in the future. 🙂

tests/step-016.sh
gabyx: True, I can look over the test and change these back, the ones which should still work. Its also tests in 117.

install.sh
gabyx: Just uncommenting, is a bit risky... Shall I better explain the sed command? Its just incremental matching lines until the last search/replace

tests/coverage.sh
rycus86: Oh right, this is never present in the production code. 👌

README.md
gabyx: You are right:
image
Hm, this fact complicats matters... Have to think if we can really handle this in the code properly...
Do you know how to split them from local paths and proper urls?

gabyx: I can just check for `.+:.+` maybe, need to take care about the @<tag> syntax we introduced

base-template.sh
gabyx: Jup, I can change that

install.sh
rycus86: We can talk about the sed later, if it has tests I'm OK with it.

My main question is why does this change keep those local paths (now converted to file URLs) working in the .shared file, when we are supposed to refuse running them now?

gabyx: Ah shit, lol, brain mistake, sorry. Jeah, its totaly stupid :-)... ups,

gabyx: We should actually move the to the local `githooks.shared`.

gabyx: If we want perfection...

install.sh
gabyx: Future: We need an SHA hash before update to check if we updated through critical commits to trigger legacy transform.

README.md
rycus86: Yeah, not sure yet what's the best way.
Have a bunch of tests to run through it? 😁
`:` might be a weak test as it would be in the front as well for things like https://
Also the branch specifier might be interesting with these like git@github.com:some/repo.git@develop

install.sh
gabyx: I think that might be pretty easy, just delete them and add them back to the --local `githooks.shared` config. But this feature is in #125. So we probably might include this changeset as well.

rycus86: Ah yeah, that could work nicely.
Any chance we could get that change for the shared config in without any of this here first?

README.md
gabyx: I think we first test for protocol, if its not there, we test for ".+:.+", if it matches we brutforce treat it as scp syntax...
of course taking into account the branch specifier...

install.sh
gabyx: I think yes. I just need to rebase, later this or next day

rycus86: Sure, no rush, thanks for all the changes!
Also sorry for being a bit slow with responses, this week is a bit busy...

gabyx: See #125 : This is to much of a burden, and we need a `dev` branch. Because I have the following plan:
- First merge #119 into `dev` 
- Then merge #125 into `dev`
- Then potentially bugfix legacy transform changes in #119
- Then make PR to introduce a `common.sh` script to outsource common functionality (really its high time!)
- Write some additiional tests for these common functions `is_local_path, is_local_url` etc...
- Finalize -> merge back to master

gabyx: So far I introduced the change in this PR as we discussed here.

base-template.sh
gabyx: Done

README.md
gabyx: This is supported now.

gabyx: Supported now.

base-template.sh
gabyx: Done

tests/step-016.sh
gabyx: Done, its only 016 which makes sense

install.sh
rycus86: > Then make PR to introduce a common.sh script to outsource common functionality (really its high time!)

I agree it's time, but that should also be it's own PR.
I'm a little bit concerned when I see 4-5 different changes in a single PR because it's 4-5 times more likely to have bugs slip through. 😔

gabyx:
Closed in Favor of #125 which contains these changes.

PR #120, opened 2020-09-09
Feature: coverage improvements

gabyx:


gabyx:
Just a pull out from the other PR....
Not yet finished

gabyx:
Coverage is back to 82%, pretty sweet :-)
Check it out :-)

gabyx:
Coverall is somehow not triggered...

gabyx:
Can you please merge this. Thanks :-)

cli.sh
```diff
@@ -293,7 +294,8 @@ git hooks disable [-r|--reset]
     fi
 
     if ! is_running_in_git_repo_root; then
-        echo "The current directory ($(pwd)) does not seem to be the root of a Git repository!"
+        echo "! The current directory ($(pwd)) does not seem" >&2
+        echo " to be the root of a Git repository!" >&2
```
rycus86, line 298: Nit for the next PR
```suggestion
        echo "  to be the root of a Git repository!" >&2
```

tests/coverage.sh
```diff
@@ -24,19 +27,30 @@ ADD examples /var/lib/githooks/examples
...
+    find /var/lib -name '*.sh' -exec sed -i -E 's|^( *)sh ~/|\1bash /home/coverage/|g' {} \\; && \\
+    # in between line
+    find /var/lib -name '*.sh' -exec sed -i -E 's|( +)sh /|\1bash /|g' {} \\; && \\
+    find /var/lib -name '*.sh' -exec sed -i -E 's|( +)sh "|\1bash "|g' {} \\; && \\
+    find /var/lib -name '*.sh' -exec sed -i -E 's|( +)sh ~/|\1bash /home/coverage/|g' {} \\; && \\
+    # git hooks alias
+    find /var/lib -name '*.sh' -exec sed -i 's|"!sh |"!bash |g' {} \\; && \\
 # Revert changed shell script filenames
     find /var/lib -name '*.sh' -exec sed -E -i "s|/var/lib/githooks/([a-z-]+)\\.bash|/var/lib/githooks/\\1.sh|g" {} \\; && \\
-# Replace the inline content with loading the source file
```
rycus86, line 38: I'd need to have a closer look, but doesn't this mean that we now count the inline content twice in the coverage?

tests/coverage.sh
```diff
@@ -24,19 +27,30 @@ ADD examples /var/lib/githooks/examples
...
-    python /var/lib/tests/replace-inline-content.py /var/lib/githooks && \\
+# Change any \`git hooks\` invocation to the shell script for better code coverage
+    find /var/lib/tests/ -name '*.sh' -exec sed -i -E 's|([^"])git +hooks|\1bash /home/coverage/.githooks/release/cli.sh|g' {} \\; && \\
+    find /var/lib/tests/ -name '*.sh' -exec sed -i -E 's|^( +)git +hooks|\1bash /home/coverage/.githooks/release/cli.sh|g' {} \\; && \\
+# Change multiline echos to line-wise echos for kcov
+    sed -i -E '/echo "$/,/^"/{  s/echo "/echo ""/ ; s/^"$/echo ""/  ;  /^\s*echo/! { s/(.*)/echo "\1"/ } }' /var/lib/githooks/cli.sh && \\
```
rycus86, line 53: Couldn't we just leave them as is and accept some drip in % coverage? How bad it is?

tests/exec-steps.sh
```diff
@@ -63,16 +63,8 @@ for STEP in /var/lib/tests/step-*.sh; do
     fi
 
     git config --global --unset init.templateDir
-    git config --global --unset githooks.shared
```
rycus86, line 66: Why did these all get deleted?

tests/coverage.sh
rycus86: Oh, we don't have inline content anymore, do we?

gabyx: There is no inline content anymore ;) :-P

gabyx: This is gone a long time ago ;-) fortunately

tests/exec-steps.sh
gabyx: Because its not needed. Since we call uninstall.sh before and that should do the reset, if anything. Maybe we could implement a test after the uninstall to make sure **no** githooks variables are set anymore.

gabyx: I implement a fail-as-early-as -possible test after the uninstall in `exec-steps.sh`: no githooks settings set ...
Check latest commits.

tests/coverage.sh
gabyx: FYI: We made huge progress I think with this code -> we moved towards git based install, we dropped all inline content, we fixed converage. This project underwent some pretty crazy changes, however thanks to the testing it all works! ;)

gabyx:
Can we merge this =), sorry for my impatience :-).

tests/coverage.sh
```diff
@@ -53,6 +67,31 @@ RUN echo "Make test gitrepo to clone from ..." && \
...
+RUN useradd -ms /bin/bash coverage
+RUN chown -R coverage:coverage /var /usr/share/git-core
+
+USER coverage
+WORKDIR /home/coverage
+RUN git config --global user.email "githook@test.com" && \
+    git config --global user.name "Githook Tests"
+
+
+## Debugging #########################################
```
gabyx, line 80: Here I left, the Debugging Code in, in case a tests fails, because kcov does not have any output -> we uncomment the kcov and run the tests and they will get copied at the end back to the host. In that way its easy to inspect the replacements.

uninstall.sh
```diff
@@ -603,6 +603,7 @@ uninstall() {
     remove_core_hooks_path
 
     # Unset global Githooks variables
+    git config --global --unset githooks.runner
```
gabyx, line 606: Bug fix, only discovered by the uninstall check in exec-tests.sh.

tests/exec-steps.sh
```diff
@@ -60,19 +60,21 @@ for STEP in /var/lib/tests/step-*.sh; do
         echo "! Uninstall failed in $STEP, output:"
         echo "$UNINSTALL_OUTPUT"
         FAILED=$((FAILED + 1))
+        break # Fail es early as possible
```
rycus86, line 63: I think it's actually better to fail at the end so we can see all the failures together rather than only the first one.
Having said that, I'm OK to undo this in the next change.

gabyx: Ok, the problem with continuation is, that it might introduce false positives and vice versa. 
But I am fine with removing the `break`.

tests/step-011.sh
```diff
@@ -2,13 +2,25 @@
 # Test:
 #   Direct template execution: test pre-commit hooks
 
+mkdir -p ~/.githooks/release && cp /var/lib/githooks/*.sh ~/.githooks/release || exit 1
```
rycus86, line 5: Can we not do this in the test setup code rather than repeated in each test?

rycus86: Actually, shouldn't the install take care of this?

gabyx: This **only** and really only for *direct* template execution. We place them there, to counteract coverage.

gabyx: Can't be done globall for all tests. 
Only for them which use *direct* template execution.

rycus86: Ah OK. 👍

gabyx:
Thx

rycus86:
Thanks a lot for this!

Issue #123, opened 2020-09-14
User feedback


gabyx:
I recently had a conversation with someone who needed assistance with setting up hooks on gitea: I thought about sharing his feedback with all of the contributors here:

>We were finally able to set up our workflow as below:
> 1) Create a repository just for server side hooks (hooks-repo)
> 2) Actual codebase repo (source-code-repo) with only client side hooks. In the server, I used git hooks shared --local <hooks-repo url> on the bare repo.
> 3) As a part of the pre-receive hook on the hooks repo, we are running the "git hooks shared update" command
> Everything worked like a breeze. This library is awesome and thanks for contributing/maintaining it. 
> Do let me know if I can contribute in any way.
>
>Thanks and Regards,
> Nithin Gunasekaran.

gabyx:
@Nithin Gunasekaran: thanks for the compliment and great that it worked out for gitea.

- I was just thinking about extending the server dok with respect to gitea (since gitea does some akward things like overwriting hooks in `.git/hooks` and basically placing the installed `githooks` runner script into a e.g. `pre-receive.d` folder which means the manager runs after the gitea hooks etc...). I will let you know when I maybe have time to do this.

- Actually we have never tested a setup to run `git hooks shared update` when executing a hook. :-P. Good that it works, it will just `git clone/pull` the shared hooks which are configured for this repo. I think you are running  `git hooks shared upadate` on the pre-receive hook in the **codebase repo** (not in the hooks repo) on the server, otherwise I dont understand how it should work :-)?

nithin001:
@gabyx, since both the repositories resided in the same server, it worked for us. The intention was to not have any server side hooks configured on the code-base-repo. It's great that you already have a PR solves that issue as well!

gabyx:
Ah, because you directly use a local path right? something like `git hooks shared --local /data/git/repositories/hooks.git`

gabyx:
@nithin001 See latest changes and `githooks.sharedHooksUpdateTriggers` you can set.

PR #124, opened 2020-09-14
Added global setting to trigger shared update by custom hook names

gabyx:
This is a PR to trigger shared hooks update by a global or local `githooks.sharedHooksUpdateTriggers` setting, which is a comma separated list (does not matter we just grep for the name) of hook names which all trigger the shared hook update additional to `post-merge` and `post-checkout 00000...`

This is related to #123 

@rycus86 : What do you think?

- [x]  Needs documentation
- ~[ ] Maybe needs a cli implementation (I think its not so important)~

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -582,8 +582,10 @@ update_shared_hooks_if_appropriate() {
     GIT_NULL_REF="0000000000000000000000000000000000000000"
 
     RUN_UPDATE="false"
+
     [ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
     [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
+    git config --get-all githooks.sharedHooksUpdateTriggers | grep -q "$HOOK_NAME" && RUN_UPDATE="true"
 
     if [ "$RUN_UPDATE" = "true" ]; then
 
--- a/install.sh
+++ b/install.sh
@@ -1590,8 +1590,7 @@ register_repo() {
 }
 
 ############################################################
-# Optionally setup shared hook repositories locally
-#   with their related Git config variables.
+# Optionally setup global shared hook repositories.
 #
 # Returns:
 #   None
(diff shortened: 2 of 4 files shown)


tests/step-018.sh
```diff
@@ -14,16 +14,18 @@ mkdir -p /tmp/shared/hooks-018.git/pre-commit &&
...
         sh ~/.githooks/release/base-template-wrapper.sh ||
     exit 1
 
-HOOK_NAME=pre-commit HOOK_FOLDER=$(pwd)/.git/hooks \
-    sh ~/.githooks/release/base-template-wrapper.sh
+OUT=$(HOOK_NAME=pre-commit HOOK_FOLDER=$(pwd)/.git/hooks \
+    sh ~/.githooks/release/base-template-wrapper.sh)
 
-if [ $? -ne 1 ]; then
+if ! echo "$OUT" | grep "Fail here"; then
```
rycus86, line 28: This just makes sure we have started the shared hook execution, but it doesn't check if it's failed as expected (by the original test). Is this change correct?

install.sh
```diff
@@ -1317,14 +1316,14 @@ setup_shared_hook_repositories() {
...
-            chmod +x ".githooks.shared.trigger" &&
-            ./.githooks.shared.trigger
-        rm -f .githooks.shared.trigger
-
-        echo "Shared hook repositories have been set up. You can change them any time by running this script again, or manually by changing the 'githooks.shared' Git config variable."
-        echo "Note: you can also list the shared hook repos per project within the .githooks/.shared file"
+
+        "$GITHOOKS_CLONE_DIR/cli.sh" shared update
+
+        echo "Shared hook repositories have been set up. You can change them any time"
```
rycus86, line 1322: /nit How did you choose the line length here?
Some of them look a bit funny. Maybe the `You can change them any time` for example could start on a new line?

base-template.sh
```diff
@@ -517,9 +517,15 @@ update_shared_hooks_if_appropriate() {
...
-    [ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
-    [ "$HOOK_NAME" = ".githooks.shared.trigger" ] && RUN_UPDATE="true"
-    [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
+
+    if [ "$HOOK_NAME" = "post-merge" ] ||
+        { [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ]; } ||
+        {
+            git config --local githooks.sharedHooksUpdateTriggers
+            git config --global githooks.sharedHooksUpdateTriggers
+        } | grep "$HOOK_NAME"; then
```
rycus86, line 526: I'd prefer to avoid `{}` where we can.
You could change this from the existing code like
```
    [ "$HOOK_NAME" = "post-merge" ] && RUN_UPDATE="true"
    [ "$HOOK_NAME" = ".githooks.shared.trigger" ] && RUN_UPDATE="true"
    [ "$HOOK_NAME" = "post-checkout" ] && [ "$1" = "$GIT_NULL_REF" ] && RUN_UPDATE="true"
    git config --local githooks.sharedHooksUpdateTriggers | grep "$HOOK_NAME" && RUN_UPDATE="true"
    git config --global githooks.sharedHooksUpdateTriggers | grep "$HOOK_NAME" && RUN_UPDATE="true"
```

rycus86:
Mostly LGTM, just left a few comments for you whenever you get a chance to have a look at them.

base-template.sh
gabyx: What does {} actually do, it gathers output together IMO?  Is it a subshell? not necessaraly right? only in a pipe `| {... }` ?

gabyx:
Rebased and added small improvments.
See the docu change.
Good to merge. Thanks =)

PR #132, opened 2020-10-06
Do not execute legacy_transform_end on single (this repo only) install 

gabyx:
Thx for merging.


rycus86:
Do we still have single install?
I thought it was already deprecated & removed?

gabyx:
Lol: We have single install, but beaware: its different meaning. It means install wrappers into this repo only, not global.

PR #133, opened 2020-10-08
Correctly detect hook template directory

gabyx:
- in the update case, we need to include a set `githooks.useCoreHooksPath` and warn and fail if we did not find a template directory after checking the default locations.

Thanks for merging.

Maybe later an additional update in 114. But not much time now.


rycus86:
I don't fully understand what's the problem here, perhaps a test might be good to make sure this doesn't regress.

gabyx:
Added the test, which fails when you checkout it.

PR #135, opened 2020-10-09
Bugfix: wrong --use-core-hookspath logic

gabyx:
This needs fixes. See test 120.sh which should not fail.
I add the fixes to this PR.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -848,8 +848,9 @@ prepare_target_template_directory() {
 
     # Up to now the directories would not have been set if
     # --use-core-hookspath is used, we set it now here.
-    if use_core_hookspath; then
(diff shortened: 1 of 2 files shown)


gabyx:
See the bugfix which fixes test 120. Thanks for merging.

gabyx:
image

install.sh
```diff
@@ -1678,8 +1679,7 @@ setup_shared_hook_repositories() {
 # Parameters:
 #   1: path for templateDir or hooksPath
 #
-# Returns:
-#   None
+# Returns: 0 if successful, 1 otherwise
```
rycus86, line 1682: We don't seem to return `1` here on failure?

PR #141, opened 2020-10-15
Bugfix/no install into existing if no previous search dir

gabyx:
Another UI Bug IMO:

```shell
git config --unset-all githooks.previousSearchDir
install.sh --non-interactive
```
directly leads to installing into existing repos (search dir).

The fix is the correct behavior, analog to the interactive prompt.

Thanks for merging.


gabyx:
Thanks!

rycus86:
> Thanks!

Thanks a lot for the fixes!

PR #147, opened 2020-10-20
Feature: Add new hooks in Git 2.28

gabyx:
Git 2.28.0 has new hooks: See changes...
- all output needs to go to stderr in base-tamplate.sh, since git might read stdin, especially for `fsmonitor-watchman`.
- add documentation

Is it ok to output everything to stderr in base-template.sh? I think yes, since that is already done by Git as I experienced (can you check this?)


rycus86:
> Is it ok to output everything to stderr in base-template.sh? 

I think I'd prefer to leave it on stdout, unless they're all exceptional messages? (update notices could perhaps also go there if that's the only other thing)

I had a quick look at the new hooks at https://git-scm.com/docs/githooks and I'm not sure if `fsmonitor-watchman` is something I can think of a good use-case for with Githooks to be honest. `post-index-change` also sounds odd for me, but I could imagine people usong that one for automation. The other two new ones look good. 👍

What do you think?

gabyx:
This is true, but this project is not dependent on the hooks basically. It should work with all of them. 🤞 
Eventhough `fsmonitor-watchman` is a funny one...

I strongly think, everything should go to stderr. I tried to contact the git mailing list about how the hooks work underneath, which descriptors get attached to what (?) it seems that git collects stdout/stderr of the hook into stderr. You can verify this in this repo by doing 
``` git commit --amend 2>/devnull```
which results  in no output.

I would really change this, also because if there are changes in the hooks such that suddenly something is parsed on stdout by Git (which is unlikely to happen I think) its better to just report stuff on stderr.

**Comment by rycus86** [comment]
> This is true, but this project is not dependent on the hooks basically. It should work with all of them. 🤞
> Eventhough `fsmonitor-watchman` is a funny one...

I think practically it's easier to say it's unsupported if no one really uses it with Githooks, than saying it's supported then dealing with odd quirks for it. 🙂

> I strongly think, everything should go to stderr. I tried to contact the git mailing list about how the hooks work underneath, which descriptors get attached to what (?) it seems that git collects stdout/stderr of the hook into stderr. You can verify this in this repo by doing
> ` git commit --amend 2>/devnull`
> which results in no output.
> 
> I would really change this, also because if there are changes in the hooks such that suddenly something is parsed on stdout by Git (which is unlikely to happen I think) its better to just report stuff on stderr.

So does this mean all the things we log to stdout currently already go to stderr? If that is the case then O don't mind if we also log to stderr if that doesn't practically change anything anyway.

**Comment by rycus86** [comment]
I'd prefer keeping stderr changes, dropping fsmonitor-watchman changes based on your comments. 🙏

**Comment by gabyx** [comment]
See my latest commit.

File: README.md [hunk 1]
```diff
@@ -72,10 +73,14 @@ The supported hooks are listed below. Refer to the [Git documentation](https://g
...
 - `post-receive`
 - `post-update`
+- `reference-transaction`
 - `push-to-checkout`
 - `pre-auto-gc`
 - `post-rewrite`
 - `sendemail-validate`
+- `post-index-change`
+
+The either sepcial hook `fsmonitor-watchman` is currently not supported. If you strongly feel otherhwise open an issue.
```
rycus86, line 83: A few typos here.
```suggestion
The `fsmonitor-watchman` hook is currently not supported. If you have a use-case for it and want to use it with this tool, please open an issue.
```

hooks/fsmonitor-watchman
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
```
rycus86, line 1: This file can now be deleted.

install.sh
```diff
@@ -665,7 +666,7 @@ check_not_deprecated_single_install() {
         echo "    You appear to have setup this repo as a single install." >&2
         echo "    The hooks in this repository are not supported anymore." >&2
         echo >&2
-        echo "    To install the latest hooks you need to reset this option"
+        echo "    To install the latest hooks you need to reset this option" >&2
```
rycus86, line 669: Nice find!

rycus86:
@gabyx Thanks, added a couple of comments.

PR #148, opened 2020-10-21
Bugfix: Fix wrong legacy check

gabyx:
- check variable to be unset

Thx for merging. Small stupid logic error :-)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -197,7 +197,7 @@ legacy_transform_after_update() {
     COMMIT_COUNT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-list --count HEAD)
     GITHOOKS_CLONE_CURRENT_COMMIT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-parse HEAD)
 
-    if [ "$COMMIT_COUNT" != "1" ] && [ -z "$INTERNAL_UPDATED_FROM_COMMIT" ]; then
+    if [ "$COMMIT_COUNT" != "1" ] && [ -z "${INTERNAL_UPDATED_FROM_COMMIT+set}" ]; then
         # If the clone dir has been updated (commit count != 1) and
         # we do not have the update commit yet (meaning we are updating from an old version )
         # we set it to the last commit where this feature (INTERNAL_UPDATED_FROM_COMMIT)


install.sh
```diff
@@ -197,7 +197,7 @@ legacy_transform_after_update() {
     COMMIT_COUNT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-list --count HEAD)
     GITHOOKS_CLONE_CURRENT_COMMIT=$(execute_git "$GITHOOKS_CLONE_DIR" rev-parse HEAD)
 
-    if [ "$COMMIT_COUNT" != "1" ] && [ -z "$INTERNAL_UPDATED_FROM_COMMIT" ]; then
+    if [ "$COMMIT_COUNT" != "1" ] && [ -z "${INTERNAL_UPDATED_FROM_COMMIT+set}" ]; then
```
rycus86, line 200: Can you explain what does this change?
Also, does this work in all Shell variants?

gabyx: This variable expansion is Posix I think.
INTERNAL_UPDATED_FROM_COMMIT is set to either "" (no update), "commitsha..." (update) or "0000..0000" (clone).
What we need to check here is: Is this variable *not* set, meanining we are before any changes where we introduces this stuf...

gabyx: > ${parameter:+word}
Use Alternative Value. If parameter is unset or null, null shall be substituted; otherwise, the expansion of word shall be substituted.

https://pubs.opengroup.org/onlinepubs/009695399/utilities/xcu_chap02.html

gabyx: Oh crap, its wrong lol

gabyx: Fixed.

gabyx:
No still swrong

gabyx:
> In the parameter expansions shown previously, use of the colon in the format shall result in a test for a parameter that is unset or null; omission of the colon shall result in a test for a parameter that is only unset. The following table summarizes the effect of the colon:

so `${aaa+thisisset}` is correct.
image

install.sh
rycus86: For me, this always makes the `-z` test false...

gabyx:
Good to merge. Now.

install.sh
rycus86: Yes, this only gives me true if I unset a variable, but doesn't match if it's set to empty string.
I still don't get why is this a problem here, but I trust you here with this code better than my understanding of it.

PR #149, opened 2020-10-27
Redirect also the output of install.sh in base-template.sh

gabyx:
Small fix. Thx for merging.

### DIFF
--- a/base-template.sh
+++ b/base-template.sh
@@ -1154,7 +1154,7 @@ execute_update() {
         return 1
     fi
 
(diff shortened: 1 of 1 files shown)


base-template.sh
```diff
@@ -460,6 +460,7 @@ execute_opt_in_checks() {
 #####################################################
 run_hook_file() {
     if [ -x "$HOOK_PATH" ]; then
+        echo "RUNNING AS EXE:"
```
rycus86, line 463: Looks like a debug message accidentally left in?

PR #151, opened 2020-10-29
Faster repository hooks :purple_heart:

gabyx:
A cherry-pick from currenty go-refactoring.

Makes these hooks way faster. I more and more god fed up that it takes too long ;-)
Tested them, works.
Thx for merging.

### DIFF
--- a/.githooks/.ignore
+++ b/.githooks/.ignore
@@ -2,3 +2,4 @@
 #   filename patterns to exclude (not paths).
 
 *.md
+*export-staged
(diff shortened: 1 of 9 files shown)


rycus86:
Can you give some more details on why is this needed? We already have STAGED_FILES set of I remember correctly, so we could just use that in the existing hooks, instead of the whole assert thing from a separate script.
Do you need that because you don't have the hooks locally when on the Go branch?

rycus86:
> No, sometimes, you want to execute a pre-commit hook single without invoking git. This ensures that it works...

OK, that makes sense. I'd have perhaps done it with a special wrapper to do the extra work of setting the staged files variable for you, but this works too I guess, just a bit more boilerplate duplicated.

```
$ sh bin/run-manually.sh .githooks/pre-commit/no-todo
```

PR #152, opened 2020-11-27
Set clone url/branch always

gabyx:


gabyx:
Small fix: Thx for merging

PR #164, opened 2023-10-20
Make script more deterministic by clearing `CDPATH`

hyperupcall:
As mentioned in [POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/cd.html), the _environment variable_ `CDPATH` chagnes the behavior of `cd`. Notably, it can cause the directory to be changed to an unexpected location. This makes the script less [deterministic](https://reproducible-builds.org/docs/deterministic-build-systems/).

My changeset causes `CDPATH` to be taken account by either unsetting it, or setting it to an empty string.


rycus86:
Hi @hyperupcall can you explain a scenario when this might cause problems? Do you perhaps have an idea for a test we could add in this repo?
The change itself sounds sensible for a bit of extra complexity, but also wondering why this wasn't necessary until now?

